PR #21, opened 2021-01-29
Authentication framework

adam-fowler:


adam-fowler:
Replaced with HummingbirdAuth library

PR #61, opened 2021-03-15
Adds 'errorResponseHandler' to HBApplication to allow HBHTTPResponseErrors to be intercepted and a custom response returned

SoftwareEngineerChris:
As promised, this is the second PR. It introduces the ability to handle a `HBHTTPResponseError` and provide a custom response, falling back to the default implementation in `HTTPResponder` using `HBHTTPResponseError` if the `errorResponseHandler` is not provided.

The reason for this PR is that I wanted to continue using standard `HBHTTPErrors` e.g. `HBHTTPError(.notFound)` at the higher-levels of the application, but have the ability to create/manipulate a custom response in-place of the ones provided by `HBHTTPResponseError` itself at a global level.

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -43,6 +43,8 @@ public final class HBApplication: HBExtensible {
     public var encoder: HBResponseEncoder
     /// decoder used by router
     public var decoder: HBRequestDecoder
+    /// Optional response handler for HBHTTPResponseError
+    public var errorResponseHandler: ((HBRequest, HBHTTPResponseError) -> EventLoopFuture<HBHTTPResponse>)?
 
     /// who provided the eventLoopGroup
     let eventLoopGroupProvider: NIOEventLoopGroupProvider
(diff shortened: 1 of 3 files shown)


adam-fowler:
Hi can I ask how you are using Hummingbird. Are you using hummingbird-core or are you just using the router elements of Hummingbird?

If it is the latter I have a minor change which will resolve this for you. The only reason I catch errors in `HTTPResponder.respond` is for a logging. Instead of converting them to an `HBHTTPResponse` at the that point I can return them as an error.

If it is the former you could use a Middleware to catch all the errors and process them eg 
```swift
struct ErrorMiddleware: HBMiddleware {
    public func apply(to request: HBRequest, next: HBResponder) -> EventLoopFuture<HBResponse> {
        return next.respond(to: request).flatMapErrorThrowing { error in
            try processError(error)
        }
    }
}
```

SoftwareEngineerChris:
Thanks Adam, I think the middleware might do the trick. I’m effectively using Hummingbird to proxy requests elsewhere that aren’t matched by the router (whilst still having some paths matched and handled by the router).

adam-fowler:
> Thanks Adam, I think the middleware might do the trick. I’m effectively using Hummingbird to proxy requests elsewhere that aren’t matched by the router (whilst still having some paths matched and handled by the router).

Make sure it is the first middleware you add to ensure you catch errors from any other middleware

SoftwareEngineerChris:
I think LambdaErrorMiddleware definitively points me in the right direction - thanks again!

SoftwareEngineerChris:
The middleware approach worked perfectly - didn’t even need to change any of the code, just shifted it into the middleware. Thanks again Adam.

PR #64, opened 2021-04-03
Update example in README

cflorion:
Hi, I know this is close to nothing but still ^^


adam-fowler (approved):
It's close to nothing but should still be fixed, thank you

PR #66, opened 2021-04-06
Init `HBApplication` with `HBHTTPServer`

adam-fowler:
Allows for overriding the default HTTP server.

This PR also includes a change to `DateCache` as we cannot guarantee which thread we are running on.


adam-fowler:
Closing for the moment, not keen on this method of doing hummingbird-ts

PR #72, opened 2021-04-21
Add HBRequestContext for extracting EventLoop, ByteBufferAllocator

adam-fowler:


adam-fowler:
Breaks websockets so this is probably going in the bin

PR #96, opened 2021-09-13
Change HBRequest/HBResponse to struct

adam-fowler:
Made `HBRequest` and `HBResponse` structs. This should make moving to a full async/await implementation a little easier as they become automatically `Sendable` as long as all their members are `Sendable`. 

For most situations, code changes related to `HBResponse` becoming a struct involves making a mutable instance of the response. 

It is slightly more complex for `HBRequest`. `HBRequest` is large enough such that converting it to a `struct` actually reduces performance. To fix this I have moved all the parameters of `HBRequest` that are read only into an internal class. There is no point copying all this data through all the response generators. Storing it in a class means only a reference to the data is passed through.

Having `HBRequest` as a struct also means some APIs that would have edited the request have had to change. You can see this in changes to [hummingbird-auth](https://github.com/hummingbird-project/hummingbird-auth/pull/10)

### DIFF
--- a/Sources/Hummingbird/Middleware/CORSMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/CORSMiddleware.swift
@@ -111,6 +111,7 @@ public struct HBCORSMiddleware: HBMiddleware {
         } else {
             // if not OPTIONS then run rest of middleware chain and add origin value at the end
             return next.respond(to: request).map { response in
+                var response = response
                 response.headers.add(name: "access-control-allow-origin", value: self.allowOrigin.value(for: request) ?? "")
                 if case .originBased = self.allowOrigin {
                     response.headers.add(name: "vary", value: "Origin")
--- a/Sources/Hummingbird/Router/ResponseGenerator.swift
+++ b/Sources/Hummingbird/Router/ResponseGenerator.swift
@@ -26,8 +26,8 @@ public protocol HBResponseGenerator {
 extension HBResponseGenerator {
     /// Generate reponse based on the request this object came from and apply request patches
     func patchedResponse(from request: HBRequest) throws -> HBResponse {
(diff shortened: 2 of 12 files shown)


mcritz:
@adam-fowler I’ll take some time to look into this. My first thought is that if this is a breaking change, then it absolutely means a major version bump.

Todo:

- [ ] Pull the branch
- [ ] Create a simple POC
- [ ] Report back results / opinions

mcritz:
@adam-fowler I think it works OK. And from an application developer perspective, I suppose its all an implementation detail. But  the `HBRequest.Internal` type feels heavy. 

What specifically is motivating the change? Are these types going to be changed to `Actor` or something?

adam-fowler:
@mcritz The main motivation behind the changes is to avoid race conditions. With `HBRequest` being a class it can be accessed/changed from multiple threads at the same time. Changing it to a struct avoids this. 

The change to a struct does have a performance cost, but by moving all the read only values into the internal class `Internal` I have managed to avoid this cost.

Much of this work has been inspired by the requirements for the `Sendable` conformance requirements for objects passed into async/await functions. See the Swift [proposal](https://github.com/apple/swift-evolution/blob/main/proposals/0302-concurrent-value-and-concurrent-closures.md). It is preparing for a full Swift concurrency implementation of Hummingbird, but it is still valid work even without that implementation as it improves stability.

adam-fowler:
> @adam-fowler I’ll take some time to look into this. My first thought is that if this is a breaking change, then it absolutely means a major version bump.
> 

Given we haven't gone 1.0 yet, we haven't committed to a interface and don't really need a major bump before releasing this. Swift concurrency may force other changes on us and I don't want to go up to v1.0 prior to that.

mcritz:
Gotcha! Cool

PR #110, opened 2021-10-30
Add Sendable conformance

adam-fowler:
Add `Sendable` conformance to Hummingbird types.
Much of this means setting `@unchecked Sendable` on types inherited from NIO
There are also a couple of places where I have added `@unchecked Sendable` on my own types because I don't have a solution yet to ensuring they are `Sendable`. ie 
- `HBEndpointResponder` where it needs to be mutable when routes are being added but post that is static. I can write code to get around this.
- `HBRequest.ResponsePatch` which is a reference object that is editable from route. 

Requiring `Sendable` is a really good way to find all your possible race conditions


ktoso (approved):
Hmmm I guess that looks good. We may keep updating how we adopt Sendable but since your types here are properly just fine it looks good. Confirming that NIO types I guess you'll be able to undo once NIO does so itself - I don't actually know what the game plan here is, will need to think or better yet ask Becca about it - there is a proposal on the forums about adopting Sendable now

adam-fowler:
too old

PR #131, opened 2022-07-07
Adjust file handling to handle loading from bundles

drekka:
I was trying to load files from a bundle of files passed to a test suite. In my server setup I had this middleware:

```swift
let configuration = HBApplication.Configuration()
let server = HBApplication(configuration: configuration)
server.middleware.add(HBFileMiddleware(Bundle.testData.filePath.absoluteString, application: server))
try server.start()
```

However whenever I requested a file from the `testData` bundle `HBFileMiddleware` failed to find it. 

I debugged through and found that the string based root folder was being set to a `file:///...` path which was not understood by the file code attempting to load the file. 

This PR:

* Converts `rootFolder` to be a file URL instead of a string.
* Updates usage of `rootFolder` based on it now being a URL. ie. `rootFolder.relativePath`.
* Removes checking for a trailing `/` which is no longer needed. 

### Testing

After the code change all Hummingbird test ran green without change.

### DIFF
--- a/Sources/HummingbirdFoundation/Cookies/Cookie.swift
+++ b/Sources/HummingbirdFoundation/Cookies/Cookie.swift
@@ -144,7 +144,7 @@ public struct HBCookie: CustomStringConvertible {
 
     /// Output cookie string
     public var description: String {
-        var output: String = "\(self.name)=\(self.value)"
+        var output = "\(self.name)=\(self.value)"
         for property in self.properties.table {
             if property.value == "" {
                 output += "; \(property.key)"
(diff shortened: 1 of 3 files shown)


adam-fowler (approved):
You need Nick Lockwood's swiftformat installed. And then you should run `./script/validate.sh`
Otherwise this is good

drekka:
Yep. I think my settings for SwiftFormat might be slightly different to yours as when I originally ran it here it changed a bunch of lines. Anyway, :-) Ran `scripts/validate.sh` which changed 3 files and committed the results.

adam-fowler:
This is released in v0.16.1

Issue #132, opened 2022-07-20
Failing to start on Mac due to in-use port fails to return 


drekka:
Not sure if I'm right about this. I have a Mac command line target that's starting a Hummingbird server.

What appears to be happening is that if the port I ask it to start on is already taken, Hummingbird fail and hangs at this point:

<img width="1423" alt="Screen Shot 2022-07-20 at 10 35 03 pm" src="https://user-images.githubusercontent.com/153837/179983313-35deeb81-37c6-4eea-82d3-2358d9cb5999.png">

This doesn't seem to occur when running the same startup code on an iOS simulator, just when running via a Mac command.

Any clues how I might resolve this?

bubudrc:
you can kill your localhost by running:

`$ lsof -n -i4TCP:8080`

PID is the second field. Then, kill that process:

`$ kill -9 PID`

[Here are more examples for different ports](https://stackoverflow.com/questions/24387451/how-can-i-kill-whatever-process-is-using-port-8080-so-that-i-can-vagrant-up)

drekka:
That's not really an option. Firstly because I don't know what's already got the port and I might need it. Second because I need to start Hummingbird from Swift. So shell scripting it's a possibility at this point.

adam-fowler:
If another process has the port, you can't bind to it, so you can't start your server. What do you expect Hummingbird to do at that point?

drekka:
I expect it to stop with an error. It does that on iOS. On OS X it simply hangs. No error, nothing.

adam-fowler:
I do remember seeing something recently about using RunLoop I'll need to do some more investigation

drekka:
Yeah. That's what I did in https://github.com/hummingbird-project/hummingbird/pull/133 but the code I have there is probably pretty tacky as it was the first thing I could think of. Did solve the issue though :-)

adam-fowler:
#135 should fix this

adam-fowler:
This is now merged into main. Can you verify this works for you

drekka:
Cool. Will get back to you.

drekka:
Hi, I've run Hummingbird on both OSX and Linux using the `main` branch and it's worked exactly as expected. Thank you. I look forwards to having it in a release.

adam-fowler:
Cheers, I should hopefully get a release out this morning

PR #133, opened 2022-07-21
Replacing future wait with run loop

drekka:
Pursuant to the discussion in https://github.com/hummingbird-project/hummingbird/issues/132 I messed about to see if I could come up with something that would allow the future to resolve and the current thread to release. 

The NIO API doco actually suggests that using `.wait()` might not be a good idea so I decided to try replacing the promise/future with run loop execution. It worked, allowing the `.start()` function to throw an error if the port was not available both iOS and OSX.

However, I'm not an expert in the the NIO code or all the depths of multithreading at that level so whilst this solutions works, it may not be the best one.

All tests were run and green with this code.

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -134,7 +134,7 @@ public final class HBApplication: HBExtensible {
         }
 
         let loop = RunLoop.current
-        while shouldKeepRunning && loop.run(mode: .default, before: Date()) {}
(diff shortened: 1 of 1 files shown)


adam-fowler:
This has been replaced with a semaphore in #135

drekka:
Cool. That wasn't something I would have thought of as I've never had need to use a semaphore in the past. Always a chance to learn new things :-)

PR #134, opened 2022-07-21
Swapping out event loop group for OS X

drekka:
This is an alternative I just figured out. It replaces the `MultiThreadedEventLoopGroup` with a `NIOTSEventLoopGroup` for OSX >= 10.14. 

A lingering thought though is will this bug still be present on Linux? Because I have someone who wants to use this on a linux machine.

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -86,11 +86,11 @@ public final class HBApplication: HBExtensible {
         self.eventLoopGroupProvider = eventLoopGroupProvider
         switch eventLoopGroupProvider {
         case .createNew:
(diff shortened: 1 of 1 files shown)


adam-fowler:
You can set the eventLoopGroup in the constructor of HBApplication. If you really wanted to use `NIOTSEventLoopGroup` on macOS you can already do that. But you'll find your server while running on Mac will service considerably less requests as `NIOTSEventLoop` is really designed for client applications. So the default should be `MultiThreadedEventLoopGroup` on macOS. It is `NIOTSEventLoopGroup` on iOS because iOS doesn't deal with POSIX threads very well.

PR #139, opened 2022-08-22
Support multiple query values with same key

adam-fowler:
Add FlatDictionary to store query values, which supports storing multiple values for the same key. Use this in HBParameters.


drekka:
Looks good to me :-)

adam-fowler:
Did a quick performance test and it is slower accessing values than a standard Dictionary, but much quicker setting values than a standard Dictionary

drekka:
Given that there's only ever a few parameters I'd be surprised it mattered :-)

PR #142, opened 2022-09-27
Use ComponentLifecycle in HBApplication

slashmo:
In my server applications I usually have two servers running, one for the public API of a service and one for the admin API (e.g. to expose a `/metrics` route), but I still want them to run in the same process/container. I'd like to be able to use `ServiceLifecycle` to pull them all together but that's currently not possible since `HBApplication` creates its own top-level lifecycle. If we were to use `ComponentLifecycle` instead, an `HBApplication` and all its lifecycle tasks could be registered as a subcomponent of a larger lifecycle, which could look like this:

```swift
let lifecycle = ServiceLifecycle()

let eventLoopGroup = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
lifecycle.registerShutdown(label: "event-loop-group", .async(eventLoopGroup.shutdownGracefully))

let publicApp = HBApplication(
    configuration: .init(
        address: .hostname(options.host, port: Int(options.publicPort)),
        serverName: "Example API",
        logLevel: options.logLevel
    ),
    eventLoopGroupProvider: .shared(eventLoopGroup)
)
lifecycle.register(publicApp.lifecycle)

let adminApp = HBApplication(
    configuration: .init(
        address: .hostname(options.host, port: Int(options.adminPort)),
        serverName: "Example API [Admin]",
        logLevel: options.logLevel
    ),
    eventLoopGroupProvider: .shared(eventLoopGroup)
)
lifecycle.register(adminApp.lifecycle)

try lifecycle.startAndWait()
```

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -97,17 +97,17 @@ public final class HBApplication: HBExtensible {
         }
 
         // create lifecycle
-        let lifecycle: LifecycleTasksContainer
+        let lifecycleTasksContainer: LifecycleTasksContainer
 
         switch serviceLifecycleProvider {
         case .shared(let parentLifecycle):
             self.lifecycle = parentLifecycle
             let componentLifecycle = ComponentLifecycle(label: self.logger.label, logger: self.logger)
-            lifecycle = componentLifecycle
+            lifecycleTasksContainer = componentLifecycle
             self.lifecycle.register(componentLifecycle)
         case .createNew:
             let serviceLifecycle = ServiceLifecycle(configuration: .init(logger: self.logger))
-            lifecycle = serviceLifecycle
+            lifecycleTasksContainer = serviceLifecycle
             self.lifecycle = serviceLifecycle
         }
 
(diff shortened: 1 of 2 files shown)


adam-fowler:
LGTM
It is API breaking but I've not committed to a v1.0 yet so not too fussed about that

adam-fowler (approved):

adam-fowler:
One issue, which is an issue with service lifecycle to some degree. ComponentLifecycle on its own does not catch interrupts like Ctrl+C.

slashmo:
> ComponentLifecycle on its own does not catch interrupts like Ctrl+C.
> @adam-fowler 

Oh, I hadn't realized that. Maybe we could fix it by adding something like a `ServiceLifecycleProvider` (inspired by `NIOEventLoopGroupProvider`) which defaults to `.createNew`, keeping the same behavior as before of creating a full `ServiceLifecycle`, but has another case `.embedded(lifecycle: ServiceLifecycle)` which would create a component life cycle and add itself to the given `ServiceLifecycle`. This could potentially also solve the API breakage if we choose to expose the full lifecycle either way. What do you think?

adam-fowler:
> Oh, I hadn't realized that. Maybe we could fix it by adding something like a `ServiceLifecycleProvider`

Yeah that would work. Might as well keep the same terminology as EventLoopGroupProvider which is `createNew` and `shared`.

slashmo:
Agreed, I'll update the PR later today.

Sources/Hummingbird/Lifecycle/ServiceLifecycleProvider.swift
```diff
@@ -0,0 +1,24 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2021-2021 the Hummingbird authors
```
adam-fowler, line 5: 2021-2022

Sources/Hummingbird/Application.swift
```diff
@@ -95,6 +95,22 @@ public final class HBApplication: HBExtensible {
         case .shared(let elg):
             self.eventLoopGroup = elg
         }
+
+        // create lifecycle
+        let lifecycle: LifecycleTasksContainer
```
adam-fowler, line 100: Can we rename this so it doesn't clash with `self.lifecycle`

adam-fowler (approved):
Looks good. 
A couple of minor nits

adam-fowler (approved):

PR #169, opened 2023-02-15
Add AsyncSequenceResponseBodyStreamer

adam-fowler:
- Using `AsyncSequenceResponseBodyStreamer`  conform `AsyncThrowingStream` and `AsyncStream` to `HBResponseGenerator`
- Also added `AsyncSequence.responseGenerator` that returns a type that wraps the `AsyncSequence` and conforms to `HBResponseGenerator`. Had to do this as I cannot conform a protocol to a new type outside of the protocol definition.


tib:
OMG this is super convenient, many thanks for adding this! 🙏

PR #179, opened 2023-03-18
Router improvements

adam-fowler:
Capture paths caught by recursive wildcard
```swift
router.get("**") { request in
    request.parameters.getCatchAll()
}
```
`GET /test/this` will return `test/this`

Wildcards with both prefix and suffix
```swift
router.get("*.jpg") { request in
}
router.get("file.*") { request in
}
```
Capture with both prefix and suffix. The parameter name is surrounded by ":" instead of just being prefixed with a ":"
```swift
router.get(":image:.jpg") { request in
    request.parameters.get("image")
}
router.get("file.:ext:") { request in
    request.parameters.get("ext")
}
```

### DIFF
--- a/Benchmarks/Sources/HBPerformance/Router.swift
+++ b/Benchmarks/Sources/HBPerformance/Router.swift
@@ -27,8 +27,8 @@ public class TrieRouterBenchmark: BenchmarkWrapper {
         self.trie.addEntry("/test/", value: "/test/")
         self.trie.addEntry("/test/one", value: "/test/one")
         self.trie.addEntry("/test/one/two", value: "/test/one/two")
-        self.trie.addEntry("/test/:value:", value: "/test/:value:")
-        self.trie.addEntry("/test/:value:/:value2:", value: "/test/:value:/:value2:")
+        self.trie.addEntry("/test/:value", value: "/test/:value")
+        self.trie.addEntry("/test/:value/:value2", value: "/test/:value:/:value2")
         self.trie.addEntry("/test2/*/*", value: "/test2/*/*")
 
         // warmup
--- a/Sources/Hummingbird/Router/Parameters.swift
+++ b/Sources/Hummingbird/Router/Parameters.swift
@@ -17,8 +17,20 @@ public struct HBParameters {
     public typealias Collection = FlatDictionary<Substring, Substring>
     internal var parameters: Collection
 
+    static let recursiveCaptureKey: Substring = ":**:"
+
     init() {
         self.parameters = .init()
+    }
+
+    init(_ values: Collection) {
+        self.parameters = values
+    }
+
+    /// Return if parameter exists
+    /// - Parameter s: parameter id
+    public func has(_ s: Substring) -> Bool {
+        return self.parameters.has(s)
     }
 
     /// Return parameter with specified id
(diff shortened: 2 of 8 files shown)


tib:
I've found an issue with the prefix / suffix capture.

```swift
router.get(":name:/:ext:") { req in
    req.parameters.get("name")! + "." + req.parameters.get("ext")!
}
```

GET /file.jpg => 404
Expected: 200, file.jpg

So we've got the following parameter cases:

  - :foo 	// Standard named parameter
  - :foo: 	// Named wildcard parameter
  - * 		// Wildcard (anything) parameter
  - ** 		// Catch all parameter

Honestly the colon thing makes my head a bit dizzy to think about parameters.

What if we use the following syntax, to make things a bit more unified:

 - hello/[name] = hello/* -> hello + catch 1 component
 - hello/[name].jpg = hello/*.jpg -> hello + catch 1 component with .jpg suffix
 - hello/[name].[ext] = hello/*.* -> hello + catch 1 component (files only)
 - */* -> catch only 2 path components -> catch 2 components
 - ** -> catch every path component -> catch all components recursively
 - hello/** -> catch all after the hello part -> hello + all components
 - [name]/** -> name + catch all after name -> name param + all components
 
I'd prefer [] characters instead of colons, it makes paths more readable.
This way you could easily capture named params in [] and use wildcards (* + **).

What do you think? Does this make any sense? 🤔

adam-fowler:
I assume you meant `:name:.:ext:`. Unfortunately you can't have prefix and suffix capture in one path component. 

I agree on the double `:`. Maybe `[]` would work. My only concern is it is completely different from full path component capture which uses the single `:`.

tib:
Yeah, so this won't work either: *.*, I just checked it. 

What if we keep the `:named` version to follow the 'standards', but allow people the [] syntax too and extend the capabilities to support capturing things within a single component? This could be a huge benefit and maybe others will follow this approach... 🤔

tib:
Also one more thing:

```
router.get("**") { req -> String? in

            print(req.parameters.getAll(":**:")) // [String]
            print(req.parameters.getAll(
                String(HBParameters.recursiveCaptureKey)
            ))
            
            print(req.uri.path) // String
            print(req.parameters.getCatchAll()) // String?
            
            // req.parameters.getCatchAll() // [String]
            
            return req.parameters.getCatchAll()
        }
```

The getCatchAll method returns an optional String, but that's not very helpful, I'd prefer to get back an array of strings (empty if nothing was catched). I can get more or less the same value using the req.uri.path property, and the only way I can access the params is the getAll(":**:") hack.

tib:
Anyway, I don't mind if we can't have multiple named captures within a component, I only needed catchAll, so this is already more than enough. Thanks for your hard work. 🙏

adam-fowler:
I ended up using `${parameter}` for parameter capture as there is prior art using this form in both shell and javascript.
You have your catchAll split into path components as well

tib:
That's a very good idea! I'll take a look at the PR shortly and come back to you with my feedbacks. Many thanks again, this is going to be super useful. 🙏

tib:
Everything looks really great, one minor suggestion: 

`public func getCatchAll() -> [String]` 

Could you return `[String]` type instead of a `Substring` array? 

I suppose most of us will have to cast the `Substring` into a `String`, so it'd make sense...

adam-fowler:
> Everything looks really great, one minor suggestion:
> 
> `public func getCatchAll() -> [String]`

You can do pretty much everything that `String` does with `Substring`. Returning an array of String's means I have to allocate memory for each String as well as the array. I would rather have the option where only the memory for the array is allocated. If you need String you can always create one from the Substring

tib:
Make sense, all right then. 👍

PR #183, opened 2023-03-22
Add support for loading `.env` files

adam-fowler:
Add static function `HBEnvironment.dotEnv()` that loads a `.env` file and creates an `HBEnvironment` from it
Add function `merging` which merges two `HBEnvironments` together.

So to get the full set of environment variables you can do the following
```swift
let env = try HBEnvironment.shared.merging(with: .dotEnv())
```


tib:
This is a very nice addition! Can't wait to test it. 👍

PR #189, opened 2023-04-28
Fix for .env parsing (#188)

### LINKED ISSUE #188: .env file parsing fails with #comments before an empty line
this test currently fails with an parser overflow error

```
    func testDotEnvCommentAndEmptyLine() throws {
        let dotenv = """
        FOO=BAR
        #BAZ=
        
        """
        let result = try HBEnvironment.parseDotEnv(dotenv)
        XCTAssertEqual(result["foo"], "BAR")
        XCTAssertEqual(result.count, 1)
    }
```

https://github.com/sliemeobn/hummingbird/blob/624afe5a405bfee8b07796b24cb7d4850f0e8405/Tests/HummingbirdTests/EnvironmentTests.swift#L117-L126

sliemeobn:
strengthened .env parsing
fixes #188

### DIFF
--- a/Sources/Hummingbird/Environment.swift
+++ b/Sources/Hummingbird/Environment.swift
@@ -157,8 +157,9 @@ public struct HBEnvironment: Sendable, Decodable, ExpressibleByDictionaryLiteral
         var parser = HBParser(dotEnv)
         var state: DotEnvParserState = .readingKey
         do {
-            parser.read(while: \.isWhitespace)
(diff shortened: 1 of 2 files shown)


adam-fowler:
You need to run `./script/validate.sh` to get formatting correct.

sliemeobn:
> You need to run `./script/validate.sh` to get formatting correct.

sorry, did not get that - should be fine now...

PR #203, opened 2023-07-11
Allow matching HBMediaType in switch statements

adam-fowler:
Allows

```swift
switch mediaType {
// will match "application/json" and "application/json; charset=utf8"
case .applicationJson:
    break
default:
    break
}

### DIFF
--- a/Tests/HummingbirdTests/HTTPTests.swift
+++ b/Tests/HummingbirdTests/HTTPTests.swift
@@ -89,6 +89,19 @@ class HTTPTests: XCTestCase {
         XCTAssert(HBMediaType(from: "audio/ogg")?.isType(.audioOgg) == true)
     }
 
+    func testMediaTypeMatching() {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: Tests/HummingbirdTests/HTTPTests.swift [hunk 1]
```diff
@@ -89,6 +89,19 @@ class HTTPTests: XCTestCase {
         XCTAssert(HBMediaType(from: "audio/ogg")?.isType(.audioOgg) == true)
     }
 
+    func testMediaTypeMatching() {
```
Joannis, line 92: We should also add `mismatching` tests

Joannis: Including `application/json` not case matching on `application/json; charset=utf8` (since the latter is more explicit). And mismatching when the type is different entirely.

Joannis (changes requested):

Joannis (approved):

PR #208, opened 2023-07-14
Error message for decoding errors

tib:
I'd like to submit a PR for returning error messages for decoding errors.
This way users could have a better understanding about what went wrong with the input.
I also wrote some test cases to verify the changes.

### DIFF
--- a/Sources/Hummingbird/Server/Request.swift
+++ b/Sources/Hummingbird/Server/Request.swift
@@ -30,7 +30,7 @@ private extension CodingKey {
     }
 }
 
-private extension [CodingKey] {
+private extension Array where Element == CodingKey {
     
     /// returns a path key using a dot character as a separator
     var pathKeyValue: String {
--- a/Sources/Hummingbird/Server/Request.swift
+++ b/Sources/Hummingbird/Server/Request.swift
@@ -20,7 +20,6 @@ import NIOCore
 import NIOHTTP1
 
 private extension CodingKey {
-
     /// returns a coding key as a path key string
     var pathKeyValue: String {
         if let value = intValue {
@@ -31,7 +30,6 @@ private extension CodingKey {
 }
 
 private extension Array where Element == CodingKey {
-    
     /// returns a path key using a dot character as a separator
     var pathKeyValue: String {
         map(\.pathKeyValue).joined(separator: ".")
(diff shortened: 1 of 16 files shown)


adam-fowler:
We still support 5.6 and 5.7 so you'll need to replace the `if let optional` syntax

Sources/Hummingbird/Server/Request.swift
```diff
@@ -100,9 +119,29 @@ public struct HBRequest: Sendable, HBSendableExtensible {
...
+        }
+        catch let DecodingError.typeMismatch(type, context)  {
+            let path = context.codingPath.pathKeyValue
+            let message = "Type mismatch for `\(path)` key, expected `\(type)` type."
+            throw HBHTTPError(.badRequest, message: message)
+        }
+        catch {
             self.logger.debug("Decode Error: \(error)")
-            throw HBHTTPError(.badRequest)
+            throw HBHTTPError(.badRequest, message: error.localizedDescription)
```
adam-fowler, line 144: You should probably use `"\(error)"` here. You get a much better description than you do from `error.localizedDescription` usually.

Sources/Hummingbird/Server/Request.swift
```diff
@@ -100,9 +119,29 @@ public struct HBRequest: Sendable, HBSendableExtensible {
...
+            let path = context.codingPath.pathKeyValue
+            let message = "Value not found for `\(path)` key."
+            throw HBHTTPError(.badRequest, message: message)
+        }
+        catch let DecodingError.typeMismatch(type, context)  {
+            let path = context.codingPath.pathKeyValue
+            let message = "Type mismatch for `\(path)` key, expected `\(type)` type."
+            throw HBHTTPError(.badRequest, message: message)
+        }
+        catch {
```
adam-fowler, line 142: Should you catch `HBHTTPError` here and then just throw it again. Otherwise if you do catch it you are throwing a `HBHTTPError` that has a message that describes the `HBHTTPError` you just caught. In actual fact to be correct you should catch all errors that conform to `HBHTTPResponseError`
```
catch let error as HBHTTPResponseError {
    throw error
}

tib: Is this how you imagined? 🤔

Sources/Hummingbird/Server/Request.swift
tib: Yeah, in this case it makes sense. It's better to display more info.

PR #214, opened 2023-07-28
Move all file middleware header parsing to threadpool

adam-fowler:


Sources/HummingbirdFoundation/Files/FileMiddleware.swift
```diff
@@ -80,121 +84,140 @@ public struct HBFileMiddleware: HBMiddleware {
...
+                    default:
+                        return request.failure(error)
+                    }
+                }
             }
         }
     }
+
+    private enum FileResult {
+        case response(HBResponse)
```
Joannis, line 218: `response` really is a `notModified` response, right?

Joannis: If so, I'd rename it to be more clear

adam-fowler: Yeah I guess, left it to be open to change but it is internal so can change at a later date if needed

adam-fowler: done

Joannis (approved):

PR #216, opened 2023-07-29
Break JobQueue handling out of HBApplication

adam-fowler:
Add `HBJobQueueHandler` which manages a job queue and a number of job queue workers
Use `HBJobQueueHandler` inside `HBApplication.JobQueueHandler`

### DIFF
--- a/Sources/HummingbirdJobs/JobQueueHandler.swift
+++ b/Sources/HummingbirdJobs/JobQueueHandler.swift
@@ -42,8 +42,47 @@ public struct HBJobQueueId: Hashable, ExpressibleByStringLiteral {
     public static var `default`: HBJobQueueId { "default" }
 }
 
+/// Object handling a single job queue
+public class HBJobQueueHandler {
+    public init(queue: HBJobQueue, numWorkers: Int, eventLoopGroup: EventLoopGroup, logger: Logger) {
+        self.queue = queue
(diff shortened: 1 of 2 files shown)


Sources/HummingbirdJobs/JobQueueHandler.swift
```diff
@@ -15,6 +15,45 @@
...
+        self.queue = queue
+        self.workers = (0..<numWorkers).map { _ in
+            HBJobQueueWorker(queue: queue, eventLoop: eventLoopGroup.next(), logger: logger)
+        }
+        self.eventLoop = eventLoopGroup.next()
+    }
+
+    /// Push Job onto queue
+    /// - Returns: Queued job information
+    @discardableResult public func enqueue(_ job: HBJob, on eventLoop: EventLoop) -> EventLoopFuture<JobIdentifier> {
```
Joannis, line 30: What does the `eventLoop` argument do?

Joannis: Is that where the (database) driver gets called from?

adam-fowler: It could be adding the job to a Redis/Fluent database

Joannis (approved):

PR #217, opened 2023-07-31
Swift service lifecycle v2

adam-fowler:
- Conform `HBApplication` to `Service` from ServiceLifecycle. `Service` requires conforming types to be `Sendable` as well. I have temporarily used `@unchecked Sendable`. Need to revisit this
- Use latest version of HummingbirdCore which include a `HBHTTPServer` conforming to `Service`.
- These changes required considerable changes to the HBXCT framework.
  -  I have remove two of the HBXCT setups (.embedded and .asyncTest) as they were too hard to maintain and replaced them with a router based one. Where requests are sent directly to the router.
  - `XCTStart` and `XCTStop` have been replaced with a single function that starts the server, runs a provided closure and then stops the server.
  - Edited all of the tests to use the new HBXCT setup
- The jobs framework is temporarily disabled. I will fix this in a later PR


Sources/Hummingbird/Configuration.swift
```diff
@@ -21,6 +21,19 @@ import Network
 extension HBApplication {
     // MARK: Configuration
 
+    /// Idle state handlder configuration
```
Joannis, line 24: Can these docs be clarified more? This is not really great.

Sources/Hummingbird/Application.swift
```diff
@@ -178,4 +124,44 @@ public final class HBApplication: HBExtensible {
...
         }
     }
+
+    public func addChannelHandler(_ handler: @autoclosure @escaping @Sendable () -> any RemovableChannelHandler) {
+        self.additionalChannelHandlers.append(handler)
+    }
+}
+
+/// Conform to `Service` from `ServiceLifecycle`.
+/// TODO: Temporarily I have added unchecked Sendable conformance to the class as Sendable
```
Joannis, line 134: Why is this unchecked Sendable necessary>

Sources/HummingbirdXCT/HBXCT.swift
```diff
@@ -48,19 +49,44 @@ struct HBXCTError: Error, Equatable {
...
 }
 
-/// Protocol for XCT framework.
-public protocol HBXCT {
-    /// Called to start testing of application
-    func start(application: HBApplication) throws
-    /// Called to stop testing of application
-    func stop(application: HBApplication)
+/// Protocol for client used by HBXCT
+public protocol HBXCTClientProtocol {
```
Joannis, line 53: Is it possible to make this conform to Sendable?

Joannis (approved):

Sources/Hummingbird/Application.swift
adam-fowler: As the comment says the protocol `Service` requires `HBApplication` to be `Sendable`. There is no way I can set it to `Sendable` now without major structural change. I will revisit this later probably when I split the applicatino into a builder type and a running type. The running type will be `Sendable` and can conform to `Service` without the `@unchecked Sendable`

Sources/HummingbirdXCT/HBXCT.swift
adam-fowler: Not easily. `HBXCTRouter.Client` holds an `HBResponder` which is not `Sendable`. Setting this to `Sendable` now would have major repercussions. I'll add a comment to revisit

PR #231, opened 2023-08-26
Date cache service

adam-fowler:


adam-fowler:
@Joannis this PR has changes based off your comments in the original PR

Sources/Hummingbird/Application.swift
```diff
@@ -54,7 +54,9 @@ public struct HBApplication: Sendable {
     // eventLoopGroup
     public let eventLoopGroup: EventLoopGroup
     // server
-    public let server: HBHTTPServer
+    internal let server: HBHTTPServer
+    // date cache service
+    public let dateCache: HBDateCache
```
Joannis, line 59: It's still `public`, is that intentional?

Joannis (approved):

Sources/Hummingbird/Application.swift
adam-fowler: damn I thought I changed that, fixed now

PR #232, opened 2023-09-01
Fix crash in tracing middleware when streaming body

adam-fowler:
You can't assume the body is available as a buffer, as it might be a stream


Sources/Hummingbird/Middleware/TracingMiddleware.swift
```diff
@@ -90,7 +90,9 @@ public struct HBTracingMiddleware: HBMiddleware {
                             switch response.body {
                             case .byteBuffer(let buffer):
                                 attributes["http.response_content_length"] = buffer.readableBytes
-                            case .stream, .empty:
+                            case .stream:
+                                attributes["http.response_content_length"] = response.headers["content-length"].first.map { Int($0) } ?? nil
```
Joannis, line 94: Why was this the lack of this attribute crashing?

adam-fowler: It is a precondition failure to access the buffer on a request that has a streamed buffer. In most cases people don't come across this because routes that aren't flagged as streaming will automatically collate the streamed buffer. 

Unfortunately request bodies are not collated by the time middleware is running.

Issue #236, opened 2023-09-18
Access-Control-Allow-Credentials both in preflight and actual request


beldis:
I tried to send a query to a Hummingbird instance with **allowOrigin: .originBase** and **allowCredentials: true**, but I get the following error (Chrome):

> Access to fetch at 'http://.../stations?application=Mobile' from origin 'http://...:3003' has been blocked by CORS policy: The value of the 'Access-Control-Allow-Credentials' header in the response is '' which must be 'true' when the request's credentials mode is 'include'

According to [CORS documentation](https://fetch.spec.whatwg.org/#cors-preflight-fetch) the _Access-Control-Allow-Credentials_ must be send both in preflight and the actual request itself, which would explain the error.

PR #237, opened 2023-09-18
Update CORSMiddleware.swift

beldis:
As mentioned in this [issue](https://github.com/hummingbird-project/hummingbird/issues/236) the Access-Control-Allow-Credentials header is missing in the actual response but send in the preflight.

Access-Control-Allow-Credentials is only set as response header when _allowCredentials_ is set to true

Changes are working on my environment and the browser are happy :)


adam-fowler (approved):
LGTM

PR #238, opened 2023-09-20
Fix SVGs in Safari

jbmorley:
SVG images weren't being displayed correctly in Safari when using `HBFileMiddleware`. Following a little investigation, it looks like Safari will only treat SVGs as images if they have the 'image/svg+xml' MIME type, whereas both Firefox and Chrome are more than happy to display them with the 'image/svg' MIME type, perhaps explaining why this wasn't seen earlier. Nginx reports SVG images as 'image/svg+xml'.

This change updates the mapping from the '.svg' extension to the 'image/svg+xml' MIME type to bring it in-line with Nginx behaviour (and presumably that of other servers).

I have verified that Safari displays SVG images correctly with this change, and Chrome and Firefox also continue to behave correctly.

### DIFF
--- a/Sources/Hummingbird/HTTP/MediaType.swift
+++ b/Sources/Hummingbird/HTTP/MediaType.swift
@@ -288,7 +288,7 @@ extension HBMediaType {
     /// Portable Network Graphics
     public static var imagePng: Self { .init(type: .image, subType: "png") }
     /// Scalable Vector Graphics (SVG)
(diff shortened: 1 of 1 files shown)


adam-fowler (approved):

adam-fowler:
Thank you for this

jbmorley:
@adam-fowler Thanks for such a quick merge! 🙏🏻

PR #243, opened 2023-10-06
Add async version of wait, add run and asyncRun

adam-fowler:
`HBApplication.wait` stalls the thread it is running on. If that thread is used by the task executor it will mean running Tasks could stall. This PR adds `asyncWait` to run application on a separate DispatchQueue.

Also added a `run` and `asyncRun` which are combinations of `start` and `wait`/`asyncWait`

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -186,7 +186,7 @@ public final class HBApplication: HBExtensible {
 
     /// Start application and wait for it to stop
     ///
-    /// Version of `run`` that can be called from asynchronous context
+    /// Version of `run` that can be called from asynchronous context
     @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
     public func asyncRun() async throws {
         try await self.onExecutionQueue { app in
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -174,7 +174,28 @@ public final class HBApplication: HBExtensible {
 
     // MARK: Methods
 
-    /// Run application
+    /// Start application and wait for it to stop
+    ///
+    /// This function can only be called from a non async context as it stalls
+    /// the current thread waiting for the application to finish
+    @available(*, noasync, message: "Use HBApplication.asyncRun instead.")
+    public func run() throws {
+        try self.start()
+        self.wait()
+    }
+
+    /// Start application and wait for it to stop
+    ///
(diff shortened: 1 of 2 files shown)


Sources/Hummingbird/Application.swift
```diff
@@ -134,7 +134,28 @@ public final class HBApplication: HBExtensible {
...
+    }
+
+    /// Start application and wait for it to stop
+    ///
+    /// Version of `run`` that can be called from asynchronous context
+    @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
+    public func asyncRun() async throws {
+        try await self.onExecutionQueue { app in
+            try app.start()
+            app.wait()
```
Joannis, line 154: This should preferably use `asyncWait()` instead, right? Code de-duplication

Sources/PerformanceTest/main.swift
```diff
@@ -54,5 +54,6 @@ app.router.get("json") { _ in
 }
 
 // run app
-try app.start()
-app.wait()
+if #available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *) {
+    try await app.asyncRun()
```
Joannis, line 58: We'll still want to have an `} else {` right?

Joannis (changes requested):

Sources/PerformanceTest/main.swift
adam-fowler: Shit that wasn't meant to go in.

Sources/Hummingbird/Application.swift
adam-fowler: Oh you mean
```swift
try app.start()
await app.asyncWait()
```

Sources/Hummingbird/Application.swift
```diff
@@ -147,11 +168,25 @@ public final class HBApplication: HBExtensible {
         try startError.map { throw $0 }
     }
 
-    /// wait while server is running
+    /// Wait until server has stopped running
+    ///
+    /// This function can only be called from a non async context as it stalls
+    /// the current thread waiting for the application to finish
+    @available(*, noasync, message: "Use HBApplication.asyncWait instead.")
```
adam-fowler, line 173: Not sure if this is a breaking change, but calling `wait` in an asynchronous context is totally borked anyway, I think we should allow this through.

adam-fowler:
@Joannis is there any issue with #242. Merging that should fix the CI for this PR

PR #247, opened 2023-10-15
HBEditedResponse: 

adam-fowler:
Build a HBResponse from optional status, headers and body conforming to HBResponseGenerator.

Not sure about the name of the type, but this is cleaner than editing the response via `request.response`


Sources/Hummingbird/Router/ResponseGenerator.swift
```diff
@@ -81,3 +73,36 @@ extension Optional: HBResponseGenerator where Wrapped: HBResponseGenerator {
         }
     }
 }
+
+public struct HBTypedResponse<Body: HBResponseGenerator>: HBResponseGenerator {
+    public var status: HTTPResponseStatus?
+    public var headers: HTTPHeaders
+    public var body: Body
```
Joannis, line 80: Is `body` a correct naming here? It's the entire response, except we potentially override properties

adam-fowler: How about 
```swift
public struct HBEditedResponse<Generator: HBResponseGenerator>: HBResponseGenerator {
    public var status: HTTPResponseStatus?
    public var headers: HTTPHeaders
    public var responseGenerator: Generator

    public init(
        status: HTTPResponseStatus? = nil,
        headers: HTTPHeaders = .init(),
        response: Generator
    ) {
        self.status = status
        self.headers = headers
        self.responseGenerator = response
    }
```

Joannis: That is a lot better to me

Joannis (approved):

PR #248, opened 2023-10-17
Generic request context

adam-fowler:
Use generics to allow the user to define their own request context type

```swift
struct MyContext: HBRequestContext {
    init(applicationContext: HBApplicationContext, channel: Channel, logger: Logger) {
        self.coreContext = .init(
            applicationContext: applicationContext, 
            channel: channel, 
            logger: logger
        )
        self.additionalData = nil
    }

    /// Field required by HBRequestContext
    public let coreContext: HBCoreRequestContext
    /// my extra data
    var additionalData: String?
}

public struct MiddlewareEditingAdditionalData: HBMiddleware {
    public typealias Context = MyContext

    public func apply(to request: HBRequest, context: Context, next: any HBResponder<Context>) -> EventLoopFuture<HBResponse> {
        var context = context
        context.additionalData = request.uri.queryParameters["test"]
        return next.respond(to: request, context: context)
    }
}

let app = HBApplicationBuilder(requestContext: MyContext.self)
app.middleware.add(MiddlewareEditingAdditionalData())
app.router.get("/test") { _, context in
    context.additionalData
}
try await app.buildAndRun()
```


adam-fowler:
@Joannis i think this is ready for review now

Package.swift
```diff
@@ -5,7 +5,7 @@ import PackageDescription
 
 let package = Package(
     name: "hummingbird",
-    platforms: [.macOS(.v10_15), .iOS(.v13), .tvOS(.v13)],
+    platforms: [.macOS(.v13), .iOS(.v16), .tvOS(.v16)],
```
Joannis, line 8: Any particular reason for the bump?

Sources/Hummingbird/Configuration.swift
```diff
@@ -18,240 +18,238 @@ import Logging
...
+        maxStreamedUploadSize: Int = 4 * 1024 * 1024,
+        maxStreamingBufferSize: Int = 1 * 1024 * 1024,
+        backlog: Int = 256,
+        reuseAddress: Bool = true,
+        tcpNoDelay: Bool = false,
+        enableHttpPipelining: Bool = true,
+        idleTimeoutConfiguration: IdleStateHandlerConfiguration? = nil,
+        threadPoolSize: Int = 2,
+        logLevel: Logger.Level? = nil,
+        noHTTPServer: Bool = false
```
Joannis, line 112: I don't like the concept and naming of `noHTTPServer`. First of all because this leads to a double-negative `noHTTPServer = false` means that the HTTP server is enabled.

But I also feel that `noHTTPServer` _really_ meas that this entire HTTPServer configuration could just be `nil`

Sources/Hummingbird/Configuration.swift
```diff
@@ -18,240 +18,238 @@ import Logging
...
-                enableHttpPipelining: enableHttpPipelining ?? self.enableHttpPipelining,
-                idleTimeoutConfiguration: idleTimeoutConfiguration ?? self.idleTimeoutConfiguration,
-                threadPoolSize: threadPoolSize ?? self.threadPoolSize,
-                logLevel: logLevel ?? self.logLevel
-            )
-        }
+    #endif
+
+    /// Create new configuration struct with updated values
+    public func with(
```
Joannis, line 200: `with` is a bit of an unclear name. Can we rename it to `modifying`? I don't have a good name at hand, just feel that `with` is a bit unclear.

Sources/Hummingbird/Server/Application+HTTPResponder.swift
```diff
@@ -55,11 +56,24 @@ extension HBApplication {
...
     }
 }
+
+extension Logger {
+    /// Create new Logger with additional metadata value
+    /// - Parameters:
+    ///   - metadataKey: Metadata key
+    ///   - value: Metadata value
+    /// - Returns: Logger
+    func with(metadataKey: String, value: MetadataValue) -> Logger {
```
Joannis, line 71: Is there a better name for `with`?

Package.swift
adam-fowler: It's needed for protocols with primary types ie HBResponder.

Sources/Hummingbird/Configuration.swift
adam-fowler: We can probably ditch this given the way the Service code works. It was mainly there to run a Jobs client without an HTTP server. Instead someone can just run the Jobs service on its own.

Sources/Hummingbird/Server/Application+HTTPResponder.swift
adam-fowler: I think in this case it is a valid name. The parameter name makes it clear what is happening. B is A with metadata.

Sources/Hummingbird/Configuration.swift
adam-fowler: Not sure of an alternative but this is slightly outside of the confines of this PR.

Sources/Hummingbird/Configuration.swift
Joannis: Let's

Joannis (approved):

PR #254, opened 2023-10-27
Merge HBApplication and HBApplicationBuilder

adam-fowler:
It does mean the router has to be built before the application, as HBRouterBuilder would be a painful to setup as Sendable. 
But the advantage of breaking out the router builder is it opens up other possibilities for other forms of route building.

```swift
let router = HBRouterBuilder()
router.middlewares.add(HBFileMiddleware())
router.get("test/") { _, context in
    return context.endpointPath
}
router.post("test2") { _, context in
    return context.endpointPath
}
let app = HBApplication(responder: router.buildResponder())
try await app.runService()
```

It does also mean the application isn't available when building routes, so data attached to application is not available (eg EventLoopGroup). But then we are trying to move away from `HBApplication` being a central source of data.


Sources/Hummingbird/Application.swift
```diff
@@ -48,60 +69,125 @@ public final class HBApplicationContext: Sendable {
...
+/// ```
+/// let app = HBApplicationBuilder()
+/// app.middleware.add(MyMiddleware())
+/// app.router.get("hello") { _ in
+///     return "hello"
+/// }
+/// try await app.buildAndRun()
+/// ```
+/// Editing the application builder setup after calling `build` will produce undefined behaviour.
+public struct HBApplication<RequestContext: HBRequestContext> {
```
Joannis, line 86: Can `HBApplication` be `Sendable` now?

Joannis (approved):

PR #255, opened 2023-10-29
Job service

adam-fowler:
Restructure Job handlers to use TaskGroup and conform to ServiceLifecycle `Service`


Sources/HummingbirdJobs/JobQueueHandler.swift
```diff
@@ -12,89 +12,94 @@
...
     }
 
-    /// Start queue workers
-    public func start() {
-        self.queue.onInit(on: self.eventLoop).whenComplete { _ in
-            self.workers.forEach {
-                $0.start()
+    public func run() async throws {
+        try await withGracefulShutdownHandler {
+            try await withThrowingTaskGroup(of: Void.self) { group in
```
Joannis, line 36: Shouldn't this be a discarding task group?

Joannis (approved):
Just one comment

Sources/HummingbirdJobs/JobQueueHandler.swift
adam-fowler: A discarding task group doesn't give you any info on when a child task has finished. So I have no way to control how many tasks are running. I think a normal task group is ok as long as you are calling next() as that'll drop finished tasks off the list. I will need to check that.

adam-fowler: Going to merge this and revisit the usage to withTaskGroup

PR #257, opened 2023-10-29
Make HBMiddleware async/await, and remove HBAsyncMiddleware

Joannis:


adam-fowler (commented):
One change to FileMiddleware
And revert persist changes to what is in async-responder

adam-fowler (approved):

PR #260, opened 2023-11-01
2.x: Async http responder

adam-fowler:


Sources/HummingbirdCore/Server/HTTPServerHandler.swift
```diff
@@ -165,6 +164,28 @@ final class HBHTTPServerHandler: ChannelDuplexHandler, RemovableChannelHandler {
                 }
             }
         }
+        /* self.responder.respond(to: request, context: context) { result in
```
Joannis, line 167: Commented code, can this be removed?

Joannis (approved):
Just one comment about comments. Otherwise fine

Sources/HummingbirdCore/Server/HTTPServerHandler.swift
adam-fowler: Shit missed that. Will remove

PR #262, opened 2023-11-01
cleanup

adam-fowler:
Remove NIOCore exports of EventLoop types
Use async/await versions of streaming functions
Remove unused EventLoop based functions


nicktrienensfuzz:
Wow this looks great, are you working to remove SwiftNIO entirely?

adam-fowler:
The plan is to make SwiftNIO a development detail. It will always be the
underlying technology for the server, but all APIs will be Swift
concurrency based.

Joannis (approved):

PR #268, opened 2023-11-06
Convert all FileIO code to async/await

adam-fowler:


Sources/HummingbirdFoundation/Files/FileIO.swift
```diff
@@ -102,105 +99,132 @@ public struct HBFileIO: Sendable {
...
+        return try await self.fileIO.write(fileHandle: handle, buffer: buffer, eventLoop: eventLoop).get()
     }
 
     /// write output of streamer to file
-    func writeFile(stream: HBByteBufferStreamer, handle: NIOFileHandle, on eventLoop: EventLoop) -> EventLoopFuture<Void> {
-        return stream.consumeAll(on: eventLoop) { buffer in
-            return self.fileIO.write(fileHandle: handle, buffer: buffer, eventLoop: eventLoop)
+    func writeFile(stream: HBByteBufferStreamer, handle: NIOFileHandle, on eventLoop: EventLoop) async throws {
+        for try await buffer in stream.sequence {
+            try await self.fileIO.write(fileHandle: handle, buffer: buffer, eventLoop: eventLoop).get()
```
Joannis, line 151: Is this an `append` operation? I thought this (over)wrote the start of the buffer

adam-fowler: This is an append operation. If you open the file with append it'll append onto the end of it, but if you open the file with create then it will clear the contents of the file and then each subsequent write will append.

I can add a test to verify if you'd like

adam-fowler: I edited `testWriteLargeFile` to ensure it writes a streamed request

Joannis (approved):

PR #270, opened 2023-11-09
HTTP 2 support

Joannis:


adam-fowler:
We should be able to avoid duplicating the work of creating the server channel.

adam-fowler:
Would #272 mean you don't need to replicate the server bootstrap creation?

adam-fowler (changes requested):
A couple of things are missing and NIO 2.62 requires a small code change
Nice idea to include the `HTTP1Channel` inside the `HTTP2Channel`. I guess we could move `handleHTTP` into HTTP1Channel if I replicate this pattern in `HTTP1WithTLSChannel`. We can do that in another PR though.

Sources/PerformanceTest/main.swift
```diff
@@ -44,6 +46,14 @@ router.get("json") { _, _ in
 
 var app = HBApplication(
     responder: router.buildResponder(),
+    channelSetup: try HTTP2Channel(
+        tlsConfiguration: .makeServerConfiguration(
+            certificateChain: [
+                .certificate(.init(file: "/Users/joannisorlandos/git/hummingbird/hummingbird/cert.pem", format: .pem))
```
adam-fowler, line 52: Path is only valid for yourself. It looks like we'll need to include AsyncHTTPClient to test HTTP2 fully. I think we should do that in a future PR though

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,97 @@
...
+        self.tlsConfiguration = tlsConfiguration
+        self.http1 = http1
+        // self.additionalChannelHandlers = additionalChannelHandlers
+        self.responder = responder
+    }
+
+    public func initialize(channel: Channel, configuration: HBServerConfiguration, logger: Logger) -> EventLoopFuture<Value> {
+        channel.eventLoop.flatSubmit {
+            do {
+                let sslContext = try NIOSSLContext(configuration: tlsConfiguration)
```
adam-fowler, line 48: Can we create this once in the initializer and store it as a member variable instead of the TLSConfiguration. Means we avoid creating it for every new child channel.

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,97 @@
...
+        responder: @escaping @Sendable (HBHTTPRequest, Channel) async throws -> HBHTTPResponse = { _, _ in throw HBHTTPError(.notImplemented) }
+    ) {
+        self.tlsConfiguration = tlsConfiguration
+        self.http1 = http1
+        // self.additionalChannelHandlers = additionalChannelHandlers
+        self.responder = responder
+    }
+
+    public func initialize(channel: Channel, configuration: HBServerConfiguration, logger: Logger) -> EventLoopFuture<Value> {
+        channel.eventLoop.flatSubmit {
```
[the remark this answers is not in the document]
Joannis: If we guarantee that, fine by me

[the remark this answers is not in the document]
adam-fowler: Yes we do. I'll add a comment to the protocol. You can add an assert as well if you'd prefer

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,97 @@
...
+        get { http1.responder }
+        set { http1.responder = newValue }
+    }
+
+    public init(
+        tlsConfiguration: TLSConfiguration,
+        http1: HTTP1Channel = HTTP1Channel(),
+        responder: @escaping @Sendable (HBHTTPRequest, Channel) async throws -> HBHTTPResponse = { _, _ in throw HBHTTPError(.notImplemented) }
+    ) {
+        self.tlsConfiguration = tlsConfiguration
```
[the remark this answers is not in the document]
Joannis: Why would we append those protocols? Shouldn't we _replace_ them? Any other protocols would just be noise. And adding them to the end would  break if one of the earlier protocols was preferred

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,97 @@
...
+
+    public func handle(value: Value, logger: Logger) async {
+        do {
+            let channel = try await value.get()
+            switch channel {
+            case .http1_1(let http1):
+                await handleHTTP(asyncChannel: http1, logger: logger)
+            case .http2((let http2, let multiplexer)):
+                try await withThrowingDiscardingTaskGroup { group in
+                    for try await client in multiplexer.inbound {
```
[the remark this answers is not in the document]
Joannis: That API is not available for `NIOHTTP2Handler.AsyncStreamMultiplexer`, since it's not a `NIOAsyncChannel`

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,97 @@
...
+                        group.addTask {
+                            await handleHTTP(asyncChannel: client, logger: logger)
+                        }
+                    }
+                }
+
+                try await http2.channel.close()
```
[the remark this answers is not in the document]
Joannis: The reason I close explicitly is because _not_ doing so would deinit `http2` implicitly closing the connection. It's simply necessary for ensuring Swift retains that Channel, nothing more.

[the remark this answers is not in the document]
Joannis: The reason I close explicitly is because _not_ doing so would deinit `http2` implicitly closing the connection. It's simply necessary for ensuring Swift retains that Channel, nothing more. But we're really not interested in reading/writing to the channel

Sources/HummingbirdHTTP2/HTTP2Channel.swift
Joannis: I kinda assumed we couldn't share it across channels or even eventloops

Joannis: Its docs clearly prove me wrong, so I've been mishandling that for a while now

Sources/PerformanceTest/main.swift
Joannis: Yeah, I didn't need to commit this you're right. I've been using `curl` to test HTTP2 and HTTP1.1 negotiation + reading

Sources/HummingbirdHTTP2/HTTP2Channel.swift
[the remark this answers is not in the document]
adam-fowler: Yeah you can replace them

Sources/HummingbirdHTTP2/HTTP2Channel.swift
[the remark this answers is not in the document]
adam-fowler: ok fair point. Might be worth pointing out to Franz this is missing

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,109 @@
...
+        tlsConfiguration: TLSConfiguration,
+        additionalChannelHandlers: @autoclosure @escaping @Sendable () -> [any RemovableChannelHandler] = [],
+        responder: @escaping @Sendable (HBHTTPRequest, Channel) async throws -> HBHTTPResponse = { _, _ in throw HBHTTPError(.notImplemented) }
+    ) throws {
+        var tlsConfiguration = tlsConfiguration
+        tlsConfiguration.applicationProtocols = NIOHTTP2SupportedALPNProtocols
+        self.sslContext = try NIOSSLContext(configuration: tlsConfiguration)
+        self.additionalChannelHandlers = additionalChannelHandlers
+        self.http1 = HTTP1Channel(additionalChannelHandlers: additionalChannelHandlers(), responder: responder)
+        self.responder = responder
```
adam-fowler, line 45: Don't need to set `self.responder` here as it is already set in the creation of `self.http1`

Sources/PerformanceTest/main.swift
```diff
@@ -14,6 +14,8 @@
 
 import Hummingbird
 import HummingbirdFoundation
+import HummingbirdHTTP2
+import NIOTLS
```
adam-fowler, line 18: You can remove the HTTP2 and TLS imports I guess

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -0,0 +1,109 @@
...
+                        }
+                    }
+                }
+
+                // Close `http2` here, because we're not interested in observing inbound/outbound here
+                // Closing it means we retain the `http2.channel`, ensuring it doesn't get deallocated
+                // until the `multiplexer` is done.
+                // Not having this will cause HTTP2 connections to close shortly after the first request
+                // is handled.
+                try await http2.channel.close()
```
adam-fowler, line 103: Your comment doesn't make sense here. Are you saying that calling a function we won't reach until all HTTP2 streams are closed stops the HTTP2 streams closing.

By the way I still think it is a good idea to close the http2 channel once we are done with it.

adam-fowler (approved):
Looks good
A couple of really minor things, plus one question

Sources/HummingbirdHTTP2/HTTP2Channel.swift
Joannis: That's exactly what happens, because if we don't refer to `http2`, its `deinit` closes the channel

Joannis: Had me scratching my head for two days

Sources/HummingbirdHTTP2/HTTP2Channel.swift
Joannis: Good catch

Sources/HummingbirdHTTP2/HTTP2Channel.swift
adam-fowler: Can you make it clear in the comment the reason it works is because you are referencing the http type. This still confuses me a little seeing as its owning type channel is still in scope. 

I think possibly the issue you had may be related to how you had everything structured previously and were adding everything to the one discarding task group. And it is no longer an issue.

Joannis: I've updated it, please review.

PR #271, opened 2023-11-09
Add missing guard keyword in Router docs

moritzsternemann:
Just noticed this small mistake in the Router guide


adam-fowler:
Thanks for that

PR #274, opened 2023-11-14
Get rid of global iso8601 formatter

adam-fowler:
It is not thread safe on Linux


tib (commented):
Isn't it going to be slow to create the formatter all the time? 🤔

adam-fowler:
> Isn't it going to be slow to create the formatter all the time? 🤔 

I checked and it isn't an issue. We can't leave it as is, because it is not thread safe and can cause crashes.

tib (approved):

Joannis (approved):

PR #275, opened 2023-11-15
Result builder router

adam-fowler:
Build router as follows
```swift
let router = HBRouter(context: HBBasicRouterContext.self) {
    HBLogRouteMiddleware()
    TracingMiddeware()
    Get("healthy") { _,_ in
        HTTPResponseStatus.ok
    }
    RouteGroup("user") {
        BasicPasswordMiddleware()
        Post(handler: signup)
    }
}
let app = HBApplication(responder: router)
```
You can return a router group from a function as follows
```swift
func todosRouteGroup<Context: HBRequestContext>() -> some HBMiddlewareProtocol<Context> {
    return RouteGroup("todos") {
        Get(handler: listTodos)
        Put(handler: putTodo)
        Get(":id", handler: getTodo) 
    }
}
let router = HBRouter(context: HBBasicRouterContext.self) {
    ...
    todosRouteGroup
    ...
}
```


adam-fowler:
This currently has an issue where it is processing middleware as it is doing route selection. So middleware inside a group where no route is matched will still run. I need to work out how to get around this.

adam-fowler:
Added creating routes with a result builder so you can apply middleware to individual routes eg
```swift
HBRouter(context: HBBasicRouterContext.self) {
    Post("login") {
        BasicPasswordMiddleware()
        login
    }
}
```

tib:
First of all, many thanks for all these updates, I really love them so far.

I only got two minor feedbacks for the current RouteBuilder setup / naming conventions:

```swift
// we don't have this right now
On(.TRACE, "foo") { _, context in
    context.resolvedEndpointPath    
}
// RouteGroup -> Group?
Group("testGroup") {
    // grouped routes & middlewares
}
```

1. I was instantly looking for an On(HTTPMethod) handler, which is non-existing at the moment.
2. The RouteGroup should be called Group IMHO, it's easier for my brain just to Group things together, I know it can be a namespace issue, but still. 😅

+1: Is this going to be the default on HB2.0? If not we should definitely create a component on top of the currently existing route system and add this as a package dependency, I'd like to be able to make a choice if possible. 🤔

What do you think about these points? 🙏

adam-fowler:
> 1. I was instantly looking for an On(HTTPMethod) handler, which is non-existing at the moment.
> 2. The RouteGroup should be called Group IMHO, it's easier for my brain just to Group things together, I know it can be a namespace issue, but still. 😅

1. You can use `Route` for `On`.
2. Naming is not defined yet. It's slightly frustrating that you can't namespace the names somehow. As result builders generate loads of really generic named top level symbols.

PR #279, opened 2023-11-18
Store `remoteAddress` in `HBCoreRequestContext`, and move responsibility for instantiation to the HTTP Server

Joannis:
This is useful when working with Lambda's, where the server does not have access to a regular `Channel`.


Joannis:
@adam-fowler is the error related to my PR? Doesn't seem like it to me

adam-fowler:
See #280. This will probably be unnecessary with that change.

PR #284, opened 2023-11-20
Support FileIO without Foundation

### LINKED ISSUE #201: 2.x: HBFileIO without Foundation
Is there a possibility we can get HBFileIO without Foundation in the future? I dug around, and I don't see why it's in the Foundation module.

https://github.com/hummingbird-project/hummingbird/blob/aa84090b472b2900003cce81208c419433c898fb/Sources/HummingbirdFoundation/Files/FileIO.swift#L22

Joannis:
Fixes #201


adam-fowler:
Are there related tests that should move with these

adam-fowler (approved):

PR #285, opened 2023-11-20
Static Let for Testing Setup

Joannis:
This allows us to add parameters down the line without breaking API


adam-fowler:
ok, given they are enums which were meant to only have one case I'm not sure it makes a difference but we can do them like this.

adam-fowler (approved):

PR #287, opened 2023-11-21
Add HBBaseRequestContext protocol

adam-fowler:
Add HBBaseRequestContext protocol that doesn't require an init using a NIO Channel
Remove HBRequestContextSource as this is no longer needed

I also found a number of places where `HBRequestContext` was being used as an existential so I fixed those at the same time.

### DIFF
--- a/Sources/Hummingbird/Server/RequestContext.swift
+++ b/Sources/Hummingbird/Server/RequestContext.swift
@@ -34,15 +34,6 @@ public struct EndpointPath: Sendable {
     private let _value: NIOLockedValueBox<String?>
 }
 
-/// Protocol for source of requests.
-///
-/// This could be the NIO Channel that created the HTTP request of a Lambda event
-public protocol RequestContextSource {
-    var eventLoop: EventLoop { get }
-    var allocator: ByteBufferAllocator { get }
-    var remoteAddress: SocketAddress? { get }
-}
-
 /// Request context values required by Hummingbird itself.
 public struct HBCoreRequestContext: Sendable {
     /// Application context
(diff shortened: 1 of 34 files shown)


Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -68,8 +59,8 @@ public struct HBCoreRequestContext: Sendable {
     public init(
         applicationContext: HBApplicationContext,
         eventLoop: EventLoop,
-        logger: Logger,
-        allocator: ByteBufferAllocator = .init()
+        allocator: ByteBufferAllocator,
```
Joannis, line 62: Any particular reason for the order change?

Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -136,6 +112,16 @@ extension HBRequestContext {
     public var id: String { self.logger[metadataKey: "hb_id"]!.description }
```
Joannis, line 112: I know this isn't in this PR, but I don't really like the way this `id` property is set up with a force unwrap.

Tests/HummingbirdTests/HandlerTests.swift
```diff
@@ -23,7 +23,7 @@ final class HandlerTests: XCTestCase {
         struct DecodeTest: HBRequestDecodable {
             let name: String
 
-            func handle(request: HBRequest, context: HBRequestContext) -> String {
+            func handle<Context: HBBaseRequestContext>(request: HBRequest, context: Context) -> String {
```
Joannis, line 26: Shouldn't these changes become `some HBBaseRequestContext`?

Joannis (approved):

Sources/Hummingbird/Server/RequestContext.swift
adam-fowler: Konrad has always made it a standard that the logger is the last parameter in a function call

Sources/Hummingbird/Server/RequestContext.swift
adam-fowler: we can discuss this after, but I'm not keen on adding another field to the context, which isn't actually used in HB. I guess now if you want an id you can implement it yourself in your own context implementation

PR #295, opened 2023-11-29
HBApplicationProtocol v2

adam-fowler:
This is version 2 of the `HBApplicationProtocol`.

Instead of making `HBApplication` a protocol, I have added the protocol `HBApplicationProtocol` and conform `HBApplication` to it. This means we can use the old method of creating an `HBApplication` but pass it around as `some HBApplicationProtocol`.

Also someone can still create their own application type conforming to `HBApplicationProtocol`

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -53,11 +53,13 @@ public final class HBApplicationContext: Sendable {
     }
 }
 
-public protocol HBApplicationProtocol: Service where Responder.Context: HBRequestContext {
+public protocol HBApplicationProtocol: Service where Context: HBRequestContext {
     /// Responder that generates a response from a requests and context
-    associatedtype Responder: HBResponder<Responder.Context>
+    associatedtype Responder: HBResponder
     /// Child Channel setup. This defaults to support HTTP1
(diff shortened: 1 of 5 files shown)


adam-fowler:
@Obbut have you got anything to add here?

Sources/Hummingbird/Application.swift
```diff
@@ -54,6 +53,119 @@ public final class HBApplicationContext: Sendable {
     }
 }
 
+public protocol HBApplicationProtocol: Service {
+    /// Context passed with HBRequest to responder
+    associatedtype Context: HBRequestContext
+    /// Responder that generates a response from a requests and context
+    associatedtype Responder: HBResponder<Context>
```
Obbut, line 60: I think this requires any `HBApplicationProtocol` implementation to explicitly specify the context type - however, `HBResponder` already defines it's own `Context` associated type, so that requirement could be dropped:

```suggestion
    /// Context passed with HBRequest to responder
    typealias Context = Responder.Context
    /// Responder that generates a response from a requests and context
    associatedtype Responder: HBResponder where Responder.Context: HBRequestContext
```

Of course the `typealias` isn't necessary per se, but might still be nice to have for ergonomics.

adam-fowler: There was a reason I did include the Context associated type, it does appear though I don't need it. It has been removed.

PR #297, opened 2023-12-02
Tests for HTTP2

adam-fowler:
Requires AsyncHTTPClient
But given I found a bug in the HTTP2 code, I think it is worthwhile


Joannis (approved):
Even without that bug we need tests :) Is there a way to test HTTP2 more thoroughly (with AHC)? For example: connection keep-alive behaviour, multiple streams per TCP connection etc

adam-fowler:
> Even without that bug we need tests :) Is there a way to test HTTP2 more thoroughly (with AHC)? For example: connection keep-alive behaviour, multiple streams per TCP connection etc

Yeah we need to extend the tests. The reason I wrote my own client was the difficultly of testing low level stuff with AHC. It hides too much about connection setup. I'll see if we can add a hook in somewhere to get a better idea of what is going on.

PR #299, opened 2023-12-02
Fix searchForIndexHtml returns wrong content-type in HBFileMiddleware

sidepelican:
When searchForIndexHtml is true, it may return `index.html`, but in that case, the `content-type` is not properly set.

### DIFF
--- a/Tests/HummingbirdFoundationTests/FileMiddlewareTests.swift
+++ b/Tests/HummingbirdFoundationTests/FileMiddlewareTests.swift
@@ -1,0 +1,52 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2021-2023 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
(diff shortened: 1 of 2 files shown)


Tests/HummingbirdFoundationTests/FileMiddlewareTests.swift
```diff
@@ -0,0 +1,52 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Hummingbird
+@testable import HummingbirdFoundation
+import HummingbirdXCT
+import XCTest
+
+final class FileMiddlewareTests: XCTestCase {
+    func testSearchForIndexHtml() throws {
```
adam-fowler, line 21: Can you move this test to FileTests.swift where all the other FileMiddleware tests are.

adam-fowler (changes requested):
Apart from one minor change looks good

Tests/HummingbirdFoundationTests/FileMiddlewareTests.swift
sidepelican: Wow, I hadn't noticed that. I've made adjustments to align with the existing tests.

adam-fowler (approved):

PR #302, opened 2023-12-05
Use MiddlewareProtocol for middleware

adam-fowler:
This is a stepping stone to including a separate result builder router. We need the middleware format to be the same between the default router and the result builder version.

I have gone with the format used by the result builder

### DIFF
--- a/Sources/Hummingbird/Middleware/CORSMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/CORSMiddleware.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Hummingbird server framework project
 //
-// Copyright (c) 2021-2021 the Hummingbird authors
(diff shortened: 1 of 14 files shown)


Joannis (commented):
I'm not opposed, but would like to see the router PR you mentioned before further evaluating/merging this. This PR in isolation feels like an unjustified change, so I'd like to see how this design change affects your other PR or repo.

adam-fowler:
> I'm not opposed, but would like to see the router PR you mentioned before further evaluating/merging this. This PR in isolation feels like an unjustified change, so I'd like to see how this design change affects your other PR or repo.

Here is the router result builder PR. The middleware types are based off the types that will come from https://github.com/swift-server/swift-middleware

Joannis (approved):

PR #304, opened 2023-12-06
Fix small code mistake in Encoding and Decoding.md

sidepelican:
Sorry this is small change, but I found small mistake of content-type in document.

And on the other hand, I think this guide should be refined with `HBMediaType`.


adam-fowler:
Do you want to change it to use `HBMediaType`?
```swift
guard let header = request.headers["content-type"].first else { throw HBHTTPError(.badRequest) }
guard let mediaType = HBMediaType(from: header) else { throw HBHTTPError(.badRequest) }
switch mediaType {
case .applicationJson:
    return try JSONDecoder().decode(type, from: request)
case .applicationUrlEncoded:
    return try URLEncodedFormDecoder().decode(type, from: request)
```

sidepelican:
I think so. 
Using `HBMediaType` seems to be more type-safe and likely to catch corner cases.

sidepelican:
Should I update this PR with the code as https://github.com/hummingbird-project/hummingbird/pull/304#issuecomment-1842446547 ?

PR #307, opened 2023-12-11
Separate HBRouter into a separate module

Joannis:
This allows Lambda users to leverage the router without having to compile NIO's HTTP code.


adam-fowler:
A first look at this implies you are not doing what you intended. HummingbirdRouter includes HummingbirdCore as a dependency which includes all the server code. You need to break the dependency on HummingbirdCore. 

The only way I saw of doing this was having a HummingbirdTypes module which included (request, response, maybe context). This would then be setup as a dependency of HummingbirdCore, HummingbirdRouter, Hummingbird. But then you are also bringing extra types in HummingbirdCore by importing HummingbirdTypes. At the moment HummingbirdCore doesn't know anything about the Context protocols.

You do have one thing in here I would like to extract the `maxUploadSize` is now part of the request context and not the application context. With that I think we can look to ditch the application context. The other parts of it are only used by the tracing middleware and are constant values that could be passed to the tracing middleware directly.

Joannis:
@adam-fowler you're right about that, I thought splitting up HBCore would be a separate PR, but that that should come in a PR before this one. I'll work on a new PR that addresses that.

The reason I didn't do that right here, is because I didn't want this PR to grow too big so it's still reviewable.

PR #308, opened 2023-12-13
Remove Application Context

adam-fowler:
Remove application context from request context
- maxUploadSize is now a requirement for the HBBaseRequestContext with a default of 2MB. If someone want to override this value they have to provide their own context type
- Added a method for adding static span attributes to tracing middleware so users can add the missing host name and port if they require

### DIFF
--- a/Sources/Hummingbird/Middleware/TracingMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/TracingMiddleware.swift
@@ -25,18 +25,17 @@ import Tracing
 @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 public struct HBTracingMiddleware<Context: HBBaseRequestContext>: HBMiddlewareProtocol {
     private let headerNamesToRecord: Set<RecordingHeader>
+    private let attributes: SpanAttributes?
 
     /// Intialize a new HBTracingMiddleware.
     ///
-    /// - Parameter recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    /// - Parameters
+    ///     - recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    ///     - parameters: A list of static parameters added to every span. These could be
     /// are being recorded.
-    public init(recordingHeaders headerNamesToRecord: some Collection<HTTPField.Name>) {
+    public init(recordingHeaders headerNamesToRecord: some Collection<HTTPField.Name> = [], attributes: SpanAttributes? = nil) {
         self.headerNamesToRecord = Set(headerNamesToRecord.map(RecordingHeader.init))
-    }
-
-    /// Intialize a new HBTracingMiddleware.
-    public init() {
-        self.init(recordingHeaders: [])
+        self.attributes = attributes
     }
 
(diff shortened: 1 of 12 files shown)


Sources/Hummingbird/Middleware/TracingMiddleware.swift
```diff
@@ -25,18 +25,17 @@ import Tracing
...
 
     /// Intialize a new HBTracingMiddleware.
     ///
-    /// - Parameter recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    /// - Parameters
+    ///     - recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    ///     - parameters: A list of static parameters added to every span. These could be
     /// are being recorded.
-    public init(recordingHeaders headerNamesToRecord: some Collection<HTTPField.Name>) {
+    public init(recordingHeaders headerNamesToRecord: some Collection<HTTPField.Name> = [], attributes: SpanAttributes? = nil) {
```
Joannis, line 36: I assume this is not related to the application context change

Sources/Hummingbird/Middleware/TracingMiddleware.swift
```diff
@@ -25,18 +25,17 @@ import Tracing
...
 public struct HBTracingMiddleware<Context: HBBaseRequestContext>: HBMiddleware {
     private let headerNamesToRecord: Set<RecordingHeader>
+    private let attributes: SpanAttributes?
 
     /// Intialize a new HBTracingMiddleware.
     ///
-    /// - Parameter recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    /// - Parameters
+    ///     - recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    ///     - parameters: A list of static parameters added to every span. These could be
```
Joannis, line 34: This doc comment seems partial

Sources/Hummingbird/Middleware/TracingMiddleware.swift
```diff
@@ -25,18 +25,17 @@ import Tracing
 @available(macOS 10.15, iOS 13.0, tvOS 13.0, watchOS 6.0, *)
 public struct HBTracingMiddleware<Context: HBBaseRequestContext>: HBMiddleware {
     private let headerNamesToRecord: Set<RecordingHeader>
+    private let attributes: SpanAttributes?
 
     /// Intialize a new HBTracingMiddleware.
     ///
-    /// - Parameter recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
+    /// - Parameters
+    ///     - recordingHeaders: A list of HTTP header names to be recorded as span attributes. By default, no headers
```
Joannis, line 33: The next line, `are being recorded`, is a line further down. So this doc comment is not right anymore

Sources/Hummingbird/Middleware/TracingMiddleware.swift
adam-fowler: This is related, in that when we remove the application context we don't have access to some of the application configuration, so cannot update some attributes eg "net.host.name" and "net.host.port". My solution was to allow the user to add some static attributes to the middleware itself.

Joannis (approved):

PR #309, opened 2023-12-13
Benchmarks 2.x

adam-fowler:
Benchmarks for HB 2.x

First run of this will fail as the benchmark code on 2.x.x branch is out of date. This needs to be merged before we see the benchmark CI working.


Benchmarks/Benchmarks/HummingbirdBenchmarks/Benchmarks.swift
```diff
@@ -12,10 +12,12 @@
...
 
-public func randomBuffer(size: Int) -> ByteBuffer {
-    var data = [UInt8](repeating: 0, count: size)
-    data = data.map { _ in UInt8.random(in: 0...255) }
-    return ByteBufferAllocator().buffer(bytes: data)
+let benchmarks = {
+    trieRouterBenchmarks()
+    routerBenchmarks()
+    // The HTTP benchmarks are so variable I'm not sure they are worthwhile running
+    //http1Benchmarks()
```
Joannis, line 22: Any idea why they're so diverse?

Joannis (approved):

Benchmarks/Benchmarks/HummingbirdBenchmarks/Benchmarks.swift
adam-fowler: I don't know if it is a timing issue causing more allocations. I would prefer if we could get these more stable as they test the whole pipeline more thoroughly

PR #310, opened 2023-12-15
Channel setup builder

adam-fowler:
Setup type for building HBChannelSetup. Cleans up API

```swift
let app = HBApplication(
    responder: router,
    channelSetup: .http1(),
    ...
)
```


Joannis (approved):
I like this change a LOT. 100% behind this

PR #313, opened 2023-12-16
2.x.x - Renaming HBChannelSetup and HBHTTPChannelSetupBuilder

adam-fowler:
- `HBChannelSetup` -> `HBChildChannel` (It doesn't just setup the child channel it handles data flowing through it)
- `HBChannelSetup.initialize` -> `HBChildChannel.setup` (initialize is too near to init)
- `HBHTTPChannelSetupBuilder` -> `HBHTTPChannelBuilder` (SetupBuilder is just weird and now HBChannelSetup has been renamed it seems sensible to remove the Setup from the name)
- `HBApplication.init(responder:channelSetup:...)` -> `HBApplication.init(responder:server:...)` (channelSetup doesn't really make sense to the user, unless they know the internals of HB)

Most of these names are internal to HB and many people won't see them, but I think it is worthwhile getting them right

### DIFF
--- a/Sources/HummingbirdCore/Server/ChildChannel.swift
+++ b/Sources/HummingbirdCore/Server/ChildChannel.swift
@@ -16,7 +16,7 @@ import Logging
 import NIOCore
 
 /// HTTPServer child channel setup protocol
-public protocol HBChannelSetup: Sendable {
+public protocol HBChildChannel: Sendable {
     associatedtype Value: Sendable
 
     /// Initialize channel
@@ -24,7 +24,7 @@ public protocol HBChannelSetup: Sendable {
     ///   - channel: channel
     ///   - childHandlers: Channel handlers to add
     ///   - configuration: server configuration
-    func initialize(channel: Channel, configuration: HBServerConfiguration, logger: Logger) -> EventLoopFuture<Value>
+    func setup(channel: Channel, configuration: HBServerConfiguration, logger: Logger) -> EventLoopFuture<Value>
 
     /// handle async channel
     func handle(value: Value, logger: Logger) async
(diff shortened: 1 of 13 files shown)


Sources/HummingbirdCore/Server/ChildChannel.swift
```diff
@@ -16,15 +16,15 @@ import Logging
 import NIOCore
 
 /// HTTPServer child channel setup protocol
-public protocol HBChannelSetup: Sendable {
+public protocol HBChildChannel: Sendable {
```
Joannis, line 19: What about `HBChannelBootstrap`?

Sources/HummingbirdHTTP2/HTTP2ChannelBuilder.swift
```diff
@@ -16,11 +16,11 @@ import HummingbirdCore
 import NIOCore
 import NIOSSL
 
-extension HBHTTPChannelSetupBuilder {
+extension HBHTTPChannelBuilder {
```
Joannis, line 19: What about `HBHTTPChannelBootstrap`?

Joannis (commented):
I agree on changing the name. But `Bootstrap` IMO more closely aligns with NIO. Regardless of what we choose, we also need to decide on if we want a technically correct name or a beginner-friendly name

Sources/HummingbirdCore/Server/ChildChannel.swift
adam-fowler: Yeah I'm not sure about that. This only deals with the setup and handling of messages on a child channel. HBServer is the bootstrap.

adam-fowler:
> I agree on changing the name. But `Bootstrap` IMO more closely aligns with NIO. Regardless of what we choose, we also need to decide on if we want a technically correct name or a beginner-friendly name

Naming very much depends on how visible the symbol is to the user. Internal symbols should always be more technically correct, public should probably be more user friendly

Joannis (approved):

PR #317, opened 2023-12-20
2.x.x - Add AsyncHTTPClient testing setup

adam-fowler:
Some testing requires a more fully featured HTTP client. This PR adds a testing framework that uses AsyncHTTPClient
```swift
try await app.test(.ahc(.https)) { client in
    client.XCTExecute(uri: "/", method: .get) { response in
    ...
```

This PR also merges HummingbirdCoreXCT into HummingbirdXCT

This PR includes #315 so it is probably best to review that first.


Sources/HummingbirdXCT/HBXCTLive.swift
```diff
@@ -76,7 +43,7 @@ final class HBXCTLive<App: HBApplicationProtocol>: HBXCTApplication {
     }
 
     init(app: App) {
-        self.timeout = .seconds(15)
+        self.timeout = .seconds(20)
```
Joannis, line 46: Any particular reason for this change?

Joannis (approved):

Sources/HummingbirdXCT/HBXCTLive.swift
adam-fowler: I had a test that had an AWS request in it which was timing out. I never got the AWS timeout because the client timed out. I should probably make this configurable.

adam-fowler: Add Issue #318

PR #321, opened 2023-12-24
2.x.x - Run services when testing using router setup

adam-fowler:
Also setup request to have authority "localhost"


Sources/HummingbirdXCT/HBXCTRouter.swift
```diff
@@ -20,25 +20,49 @@ import Logging
...
+        self.services = app.services
         self.logger = app.logger
     }
 
     /// Run test
     func run<Value>(_ test: @escaping @Sendable (HBXCTClientProtocol) async throws -> Value) async throws -> Value {
         let client = Client(eventLoopGroup: self.eventLoopGroup, responder: self.responder, logger: self.logger)
-        let value = try await test(client)
-        return value
+        if self.services.count > 0 {
```
Joannis, line 42: If we handle the `else` (isEmpty) clause first, we can de-nest this block

Joannis (approved):
LGTM, but maybe re-organise the branching of this code

PR #326, opened 2023-12-29
Add RequestID to generate unique Identifier for each request

adam-fowler:
Previously this was just an incrementing `Int`, but it was pointed out that ids across instances of an Hummingbird app would clash as they all started at zero. This is now two integers one identifying the instance and the other an incrementing `UInt64` started at a random point in its range. This has all been encapsulated in a `RequestID` type.

### DIFF
--- a/Sources/Hummingbird/Server/RequestID.swift
+++ b/Sources/Hummingbird/Server/RequestID.swift
@@ -17,15 +17,13 @@ import Atomics
 /// Generate Unique ID for each request
 struct RequestID: CustomStringConvertible {
     let low: UInt64
-    let high: UInt64
 
     init() {
         self.low = Self.globalRequestID.loadThenWrappingIncrement(by: 1, ordering: .relaxed)
-        self.high = Self.instanceIdentifier
     }
 
     var description: String {
-        String(self.high, radix: 16) + self.formatAsHexWithLeadingZeros(self.low)
+        String(Self.high, radix: 16) + self.formatAsHexWithLeadingZeros(self.low)
     }
 
     func formatAsHexWithLeadingZeros(_ value: UInt64) -> String {
@@ -37,6 +35,6 @@ struct RequestID: CustomStringConvertible {
         }
     }
 
-    private static let instanceIdentifier = UInt64.random(in: .min ... .max)
+    private static let high = UInt64.random(in: .min ... .max)
     private static let globalRequestID = ManagedAtomic<UInt64>(UInt64.random(in: .min ... .max))
 }
--- a/Sources/Hummingbird/Server/RequestID.swift
+++ b/Sources/Hummingbird/Server/RequestID.swift
@@ -23,7 +23,7 @@ struct RequestID: CustomStringConvertible {
     }
 
     var description: String {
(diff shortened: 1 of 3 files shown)


Joannis (approved):
I agree with this change, at the very least for 2.x. Though I'm wondering to what extend users depend on the nature of this having been an integer

adam-fowler:
> I agree with this change, at the very least for 2.x. Though I'm wondering to what extend users depend on the nature of this having been an integer

This was a request from someone using v1. So I'd prefer to merge into v1.

Hmmm, I don't know about the integer thing, The fact it was an integer is really a side effect of how the id was generated. I can be explicit in the release notes about the id change and how it might affect users who depended on it being an integer that fitted into a Int64.

Sources/Hummingbird/Server/RequestID.swift
```diff
@@ -0,0 +1,40 @@
...
+struct RequestID: CustomStringConvertible {
+    let low: UInt64
+
+    init() {
+        self.low = Self.globalRequestID.loadThenWrappingIncrement(by: 1, ordering: .relaxed)
+    }
+
+    var description: String {
+        String(Self.high, radix: 16) + self.formatAsHexWithLeadingZeros(self.low)
+    }
```
MahdiBM, line 27: * It's certainly cleaner this way, but it still feels unnecessary to transform the integer to a `String` once per `RequestID.description` call, considering `.description` will be common due to the logging statement.
* Doesn't this hex description (partially) defeat the point of having a nice easily-human-readable integer as the request's id? At that point a random UUID could just be the simpler solution although i assume it's somewhat more costly to create.

MahdiBM (commented):
I realize there might be someone somewhere somehow counting on the ID being an Int, but that's quit unlikely as the id is only used in the log statement (unlike Vapor which exposes the request id to the public, for reasons that i think are legitimate but i still don't remember).

Sources/Hummingbird/Server/RequestID.swift
adam-fowler: Good point about transforming `high` repeatedly. 

I'm not using UUID because one it requires Foundation and two it is slower to generate. I reduced it to hex to shorten the id and the first 16 chars end up being a process id. Anyway not sure a 128bit integer is human readable.

adam-fowler:
> I realize there might be someone somewhere somehow counting on the ID being an Int, but that's quit unlikely as the id is only used in the log statement (unlike Vapor which exposes the request id to the public, for reasons that i think are legitimate but i still don't remember).

I think @Joannis requested that. He requested the same of HB and a `String` is returned as an id. If someone is converting that back to an integer I would want to discourage that anyway.

Sources/Hummingbird/Server/RequestID.swift
MahdiBM: Hmm fair enough. I now think it's fine being a hex 🤔 i was thinking someone might rather read/have it as a decimal number, but I can't come up with enough reasons why that should be important. They can always turn the hex back to an integer trivially if it's that important 🤔

PR #329, opened 2024-01-02
2.x.x - Add HBApplication.init(router:...)

adam-fowler:
Add `HBApplication(router:server:configuration:eventLoopGroupProvider:)`. Removes requirement for calling `buildResponder` therefore cleaner API

Need to make HBRouterResponder public and explicitly return it from `buildResponder` to get this to work


adam-fowler:
> Definitely a great improvement to UX

Is that an approval then 😄 ?

PR #330, opened 2024-01-05
2.x.x - Trailer headers

adam-fowler:
Trailer headers are output as the return value from the response body writer

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBody.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBody.swift
@@ -28,14 +28,14 @@ public struct HBResponseBody: Sendable {
     /// - Parameters:
     ///   - contentLength: Optional length of body
     ///   - write: closure provided with `writer` type that can be used to write to response body
-    public init(contentLength: Int? = nil, _ write: @Sendable @escaping (any HBResponseBodyWriter) async throws -> HTTPFields?) {
-        self.write = write
+    public init(contentLength: Int? = nil, _ write: @Sendable @escaping (any HBResponseBodyWriter) async throws -> Void) {
+        self.write = { try await write($0); return nil }
         self.contentLength = contentLength
     }
 
     /// Initialise empty HBResponseBody
     public init() {
-        self.init(contentLength: 0) { _ in return nil }
+        self.init(contentLength: 0) { _ in }
     }
 
     /// Initialise HBResponseBody that contains a single ByteBuffer
@@ -43,7 +43,6 @@ public struct HBResponseBody: Sendable {
     public init(byteBuffer: ByteBuffer) {
         self.init(contentLength: byteBuffer.readableBytes) { writer in
             try await writer.write(byteBuffer)
-            return nil
         }
     }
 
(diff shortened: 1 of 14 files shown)


Joannis:
Many users won't use the trailing headers, does it make sense to support an additional closure that's (still) defines as `-> Void` for users that don't need trailing headers? We could then make the trailing headers return a non-optional, so it's clearly distinct.

adam-fowler:
> Many users won't use the trailing headers, does it make sense to support an additional closure that's (still) defines as `-> Void` for users that don't need trailing headers? We could then make the trailing headers return a non-optional, so it's clearly distinct.

I tried having two different `init` one with a handler that returned `Void` and one with a handler that return `HTTPFields` but it confuses the compiler. If you have a closure where it isn't obvious what the return value you get an error. I think I have a solution though checkout the last commit.

Sources/HummingbirdCore/Response/ResponseBody.swift
```diff
@@ -50,6 +53,30 @@ public struct HBResponseBody: Sendable {
...
+    /// Create HBResponseBody that returns trailing headers from its closure once all the 
+    /// body parts have been written 
+    /// - Parameters:
+    ///   - contentLength: Optional length of body
+    ///   - write: closure provided with `writer` type that can be used to write to response body
+    ///         trailing headers are returned from the closure after all the body parts have been
+    ///         written
+    public static func withTrailingHeaders(
+        contentLength: Int? = nil, 
+        _ write: @Sendable @escaping (any HBResponseBodyWriter) async throws -> HTTPFields?
```
Joannis, line 69: I like this a lot better, now one more idea to play with: Can we make the `any HBResponseBodyWriter` an `inout`? This would make it harder to incorrectly wield this API.

adam-fowler: I'm not sure what you are getting at there. What do we gain from  making `any HBResponseBodyWriter` `inout`. That closure is called by `HTTPChannelHandler` and any Middleware that want to process response bodies. ie very few people will be calling it.

Are you seeing a situation where you would want to the write function to mutate the `HBResponseBodyWriter`?

Joannis: No, I don't see a situation where we'd want to mutate. I see a risk in people trying to escape that HBResponseBodyWriter out of this closure

Joannis (approved):

Sources/HummingbirdCore/Response/ResponseBody.swift
Joannis: I don't want to hold up this PR for this topic though, since it's not caused by this PR

adam-fowler: ok I see, by making it inout you make it harder but it could still be done. I could make `~Copyable` I don't know if that'd help

adam-fowler: Forgot that HBResponseBodyWriter is a protocol not a struct

adam-fowler: Adding an issue to chase this up

PR #331, opened 2024-01-05
2.x.x - Change parameter capture syntax to same as OpenAPI

adam-fowler:
In v1.0 to support prefix and suffix URI parameters we added a new syntax for defining extracted parameters `${parameter}`. Since then the OpenAPI runtime has come on line and it seems more sensible to duplicate the syntax OpenAPI has `{parameter}`. ie remove the `$`

### DIFF
--- a/Sources/Hummingbird/Router/RouterPath.swift
+++ b/Sources/Hummingbird/Router/RouterPath.swift
@@ -30,11 +30,11 @@ public struct RouterPath: Sendable, ExpressibleByStringLiteral, CustomStringConv
             case .path(let path):
                 return String(path)
             case .capture(let parameter):
-                return "${\(parameter)}"
+                return "{\(parameter)}"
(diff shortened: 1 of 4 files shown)


Joannis (commented):
I don't mind this change, it makes the ecosystem more consistent. My main worry is about needing a clear note in docs for migrating older users

adam-fowler:
> I don't mind this change, it makes the ecosystem more consistent. My main worry is about needing a clear note in docs for migrating older users

I understand your concern. But being consistent is probably a good thing and if there is any time to change this kind of thing it is a major release. We can make sure to be clear about this change in the release notes and any migration docs.

adam-fowler:
@tib given you were involved in the original decision regarding prefix and suffix capture parameters. How do you feel about changing it from `${param}` to `{param}` to be more consistent with open api.

PR #332, opened 2024-01-05
HBRouteHandler cleanup

adam-fowler:
- `handle` method only gets the context as it is assumed you have extracted all you need from the request in `init`
- Removed `HBRequestDecodable` as soon as that was released I hated it. 
- correct comment

Not sure if I'll keep the route handlers as it is hard to do dependency injection with them. But for the moment I've cleaned up everything a little

I also changed `HBRequest.decode(as:using)` to `HBRequest.decode(as:context)`. I don't use `using` to pass the context around anyway else. It is always context. This change went in here as it clashed with the rest of this PR


Joannis:
@adam-fowler I like removing `HBRequestDecodable`, and understand removing the request from `handle`. But since `init` also gets the context, shouldn't we remove that from `handle` as well?

adam-fowler:
> @adam-fowler I like removing `HBRequestDecodable`, and understand removing the request from `handle`. But since `init` also gets the context, shouldn't we remove that from `handle` as well?

Then people would have to store things like the logger and allocator in the route handler

Joannis (approved):
I'm not sure what to think of the remaining issue I've brought up, but this improvement is already a welcome one. Should we make an discussion topic/issue of my comment?

adam-fowler:
> I'm not sure what to think of the remaining issue I've brought up, but this improvement is already a welcome one. Should we make a discussion topic/issue of my comment?

Yeah add an issue

PR #333, opened 2024-01-05
HBRequest.collateBody

adam-fowler:
There is an issue with the current API for read request bodies. As it is a stream of ByteBuffers you can read it only once. I had to provide a solution for someone who wants to read the request body in middleware (never a great idea) but still have it available for route handlers.

With the current setup it is very difficult to get the contents of the request a second time if it has been collated into one buffer previously.

The changes I've made change the `HBRequestBody.collate(maxSize:)`  to return a ByteBuffer instead of a new `HBRequestBody` and this should be the function people use instead of `AsyncSequence.collect(upTo:)`. I've also added a helper function that collates the request body in place in a request which can be used by Middleware if they need it.

### DIFF
--- a/Sources/Hummingbird/Server/Request.swift
+++ b/Sources/Hummingbird/Server/Request.swift
@@ -15,6 +15,20 @@
 import HummingbirdCore
 
 extension HBRequest {
+    /// Collapse body into one ByteBuffer.
+    ///
+    /// This will store the collated ByteBuffer back into the request so is a mutating method. If
+    /// you don't need to store the collated ByteBuffer on the request then use
+    /// `request.body.collate(maxSize:)`.
+    ///
+    /// - Parameter context: request context
+    /// - Returns: Collated body
+    public mutating func collateBody(context: some HBBaseRequestContext) async throws -> ByteBuffer {
+        let byteBuffer = try await self.body.collate(maxSize: context.maxUploadSize)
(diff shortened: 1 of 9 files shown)


Sources/Hummingbird/Server/Request.swift
```diff
@@ -15,6 +15,20 @@
...
+    /// Collapse body into one ByteBuffer.
+    ///
+    /// This will store the collated ByteBuffer back into the request so is a mutating method. If
+    /// you don't need to store the collated ByteBuffer on the request then use
+    /// `request.body.collate(maxSize:)`.
+    ///
+    /// - Parameter context: request context
+    /// - Returns: Collated body
+    public mutating func collateBody(context: some HBBaseRequestContext) async throws -> ByteBuffer {
+        let byteBuffer = try await self.body.collate(maxSize: context.maxUploadSize)
```
Joannis, line 27: One thing I've started wondering; in most cases we'll know the body's size through content-length. Should we leverage that information in any way?

Joannis (approved):
I like this change, but I do wonder about a related issue not caused by your change

Sources/Hummingbird/Server/Request.swift
adam-fowler: When you are collating the request body you could use it to compare the actual body size with the content-length and return a badRequest if they differ?

PR #334, opened 2024-01-05
Provide logger to HBApplication

adam-fowler:
Make sure we can use same Logger across all services instead of HBApplication crearing its own

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -201,7 +201,7 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
         if let logger {
             self.logger = logger
         } else {
-            var logger = Logger(label: configuration.serverName ?? "HummingBird")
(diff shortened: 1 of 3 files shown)


Sources/Hummingbird/Application.swift
```diff
@@ -184,16 +184,27 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
...
+        logger: Logger? = nil
     ) {
-        var logger = Logger(label: configuration.serverName ?? "HummingBird")
-        logger.logLevel = configuration.logLevel
-        self.logger = logger
-
+        if let logger {
+            self.logger = logger
+        } else {
+            var logger = Logger(label: configuration.serverName ?? "HummingBird")
```
MahdiBM, line 204: Nitpick:
```suggestion
            var logger = Logger(label: configuration.serverName ?? "Hummingbird")
```

Joannis (approved):

adam-fowler:
Cannot work out why CI is failing. It seems to be using the wrong code

PR #337, opened 2024-01-07
2.x.x - Don't shutdown server in HBApplication

adam-fowler:
The service group will do that for us

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -122,11 +122,7 @@ extension HBApplicationProtocol {
                 configuration: .init(services: services, logger: self.logger)
             )
             try await serviceGroup.run()
-        } onGracefulShutdown: {
-            Task {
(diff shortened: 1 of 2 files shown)


Sources/Hummingbird/Application.swift
```diff
@@ -122,11 +122,7 @@ extension HBApplicationProtocol {
                 configuration: .init(services: services, logger: self.logger)
             )
             try await serviceGroup.run()
-        } onGracefulShutdown: {
-            Task {
-                try await server.shutdownGracefully()
-            }
-        }
+        } onGracefulShutdown: {}
```
Joannis, line 125: I might misinterpret this, but is there any point in the graceful shutdown handler block now?

adam-fowler: You're right we don't need it

Joannis (approved):

PR #338, opened 2024-01-07
Add HBApplication.addServices(_ services: [any Service])

adam-fowler:


Sources/Hummingbird/Application.swift
```diff
@@ -246,6 +246,12 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
         self.services.append(contentsOf: services)
     }
 
+    ///  Add service to be managed by application ServiceGroup
+    /// - Parameter services: list of services to be added
+    public mutating func addServices(_ services: [any Service]) {
```
Joannis, line 251: Since `self.services` is already public, I don't really see the benefits of adding this function.

adam-fowler:
Hadn't realised it was public

PR #340, opened 2024-01-08
2.x.x - Run process before server start

adam-fowler:
While writing the Todos tutorial I came across an issue with database setup. There are situations where you need to do pre-processing of your database before starting the server eg running migrations. Unfortunately the new PostgresClient requires that you have a background process running, not currently managed by ServiceLifecycle but I assume it will be eventually. So you can't do the database preprocessing before the server has started because it requires the PostgresClient service to be running, which would also kick off the server service. 

This PR adds a way to add processes to be run before the server has started but while all other services are running. 

```swift
let db = DatabaseClient()
let app = HBApplication()
app.addService(db)
add.runBeforeServerStart {
    db.migrate()
}
```

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -184,7 +184,7 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
     public let server: HBHTTPChannelBuilder<ChildChannel>
     /// services attached to the application.
     public var services: [any Service]
-    /// processed to be run before server is started
+    /// Processes to be run before server is started
     public private(set) var processesRunBeforeServerStart: [@Sendable () async throws -> Void]
 
     // MARK: Initialization
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Hummingbird server framework project
 //
-// Copyright (c) 2021-2023 the Hummingbird authors
+// Copyright (c) 2021-2024 the Hummingbird authors
 // Licensed under Apache License v2.0
 //
 // See LICENSE.txt for license information
@@ -64,6 +64,9 @@ public protocol HBApplicationProtocol: Service where Context: HBRequestContext {
     @Sendable func onServerRunning(_ channel: Channel) async
     /// services attached to the application.
     var services: [any Service] { get }
+    /// Array of processes run before we kick off the server. These tend to be processes that need
+    /// other services running but need to be run before the server is setup
+    var processesRunBeforeServerStart: [@Sendable () async throws -> Void] { get }
 }
 
 extension HBApplicationProtocol {
@@ -82,6 +85,8 @@ extension HBApplicationProtocol {
     public func onServerRunning(_: Channel) async {}
     /// Default to no extra services attached to the application.
     public var services: [any Service] { [] }
+    /// Default to no processes being run before the server is setup
+    public var processesRunBeforeServerStart: [@Sendable () async throws -> Void] { [] }
 }
 
 /// Conform to `Service` from `ServiceLifecycle`.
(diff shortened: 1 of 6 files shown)


Tests/HummingbirdTests/ApplicationTests.swift
```diff
@@ -531,6 +531,22 @@ final class ApplicationTests: XCTestCase {
         XCTAssertEqual(MyService.shutdown.load(ordering: .relaxed), true)
     }
 
+    func testRunBeforeServer() async throws {
+        let runBeforeServer = ManagedAtomic(false)
+        let router = HBRouter()
+        var app = HBApplication(
+            responder: router.buildResponder()
+        )
```
adam-fowler, line 539: Once #339 is merged I can add a check to see if the atomic has been set by then

Sources/Hummingbird/Utils/GracefulShutdownWaiter.swift
```diff
@@ -2,7 +2,7 @@
 //
 // This source file is part of the Hummingbird server framework project
 //
-// Copyright (c) 2021-2021 the Hummingbird authors
+// Copyright (c) 2021-2023 the Hummingbird authors
```
Joannis, line 5: 2024

Sources/Hummingbird/Utils/PrecursorService.swift
```diff
@@ -0,0 +1,31 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2021-2023 the Hummingbird authors
```
Joannis, line 5: 2024

Sources/HummingbirdXCT/TestApplication.swift
```diff
@@ -35,20 +35,23 @@ struct TestApplication<BaseApp: HBApplicationProtocol>: HBApplicationProtocol, S
...
     @Sendable func onServerRunning(_ channel: Channel) async {
         await self.portPromise.complete(channel.localAddress!.port!)
     }
 
-    /// services attached to the application.
+    /// Services attached to the application.
     var services: [any Service] { self.base.services }
 
+    /// Processes run before server start
+    public var processesRunBeforeServerStart: [@Sendable () async throws -> Void] { self.base.processesRunBeforeServerStart }
```
Joannis, line 53: I don't like this variable name, but I don't know a better one.

Tests/HummingbirdTests/ApplicationTests.swift
Joannis: #339 can be merged

Joannis (approved):
Please do fix the test, but otherwise approved

Sources/Hummingbird/Application.swift
```diff
@@ -180,6 +190,8 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
     public let server: HBHTTPChannelBuilder<ChildChannel>
     /// services attached to the application.
     public var services: [any Service]
+    /// processed to be run before server is started
```
MahdiBM, line 193:
```suggestion
    /// processes to be run before server is started
```

Sources/HummingbirdXCT/TestApplication.swift
MahdiBM: I guess we could use `pre`. Like `preStartProcesses`. A little bit more vague than `processesRunBeforeServerStart`, but should do the job.

In that case, there is also the `runBeforeServerStart` function that needs to have a change of name to something like `preStartRun`. Not sure if these are the best options, but they sound good to me.

adam-fowler: Yeah it's not a great variable name, but I don't think @MahdiBM your suggestions are much better especially the function rename. `preStartRun` doesn't mean a great deal.

adam-fowler: Gonna add a Bikeshedding issue with all the symbol names that need revisiting and add this to it

adam-fowler: #343

PR #341, opened 2024-01-10
Make HBXCTResponse.body non optional

adam-fowler:
Remove requirement for tests to check if response body is non-optional.


Sources/HummingbirdXCT/HBXCTRouter.swift
```diff
@@ -22,7 +22,7 @@ import NIOCore
 import NIOPosix
 import ServiceLifecycle
 
-/// Test sending values to requests to router. This does not setup a live server
+/// Test sending equests directly to router. This does not setup a live server
```
Joannis, line 25:
```suggestion
/// Test sending requests directly to router. This does not setup a live server
```

Joannis (approved):

Issue #343, opened 2024-01-11
2.x.x - Bikeshedding


adam-fowler:
Here are a list of symbol names, I'm not overly happy with but couldn't think of a better one at the time...

- [ ] `HBApplicationProtocol.processesRunBeforeServerStart`: An array of closures we run before starting the HBServer service. 
- [ ] `HTTPChannelHandler`: Protocol for child channel that can handle HTTP1.1 packets.

Diggory:
how about `prestartProcesses` or `prestartActions`?

adam-fowler:
> how about `prestartProcesses` or `prestartActions`?

It has to be clear these are processes that will be run before we start the server.

Joannis:
Personally I think we're fine with these names

PR #344, opened 2024-01-11
Add HBResponderBuilder

adam-fowler:
Added this to fix a weird API thing where HBRouterBuilder could not be used as a paramter to `HBApplication(router:...)`

All routers should conform to this `protocol` and provide a `buildResponder` method

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -232,14 +232,14 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
     ///   - onServerRunning: Function called once the server is running
     ///   - eventLoopGroupProvider: Where to get our EventLoopGroup
     ///   - logger: Logger application uses
-    public init<Context>(
-        router: HBRouter<Context>,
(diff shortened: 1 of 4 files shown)


Sources/Hummingbird/Application.swift
```diff
@@ -232,14 +232,14 @@ public struct HBApplication<Responder: HBResponder, ChildChannel: HBChildChannel
     ///   - onServerRunning: Function called once the server is running
     ///   - eventLoopGroupProvider: Where to get our EventLoopGroup
     ///   - logger: Logger application uses
-    public init<Context>(
-        router: HBRouter<Context>,
+    public init<ResponderBuilder: HBResponderBuilder>(
+        router: ResponderBuilder,
```
Joannis, line 236: The only remark is that I don't know how correct this argument's name is

Joannis (approved):

Sources/Hummingbird/Application.swift
adam-fowler: Yeah maybe another to add to #343

The idea of adding the router parameter was to make it clear we were adding a router but in theory it could be anything that creates a responder.

PR #345, opened 2024-01-16
Routerbuilder docs

adam-fowler:
- Added documentation for HummingbirdRouter.
- Removed unused symbols
- Hid internal (but public) symbols
- Moved some code around to clean up module

Minor change to `RouteGroup.init`. I found I don't need the context parameter (compiler is clever enough to work it out) and I removed the default value for the path.

### DIFF
--- a/Tests/HummingbirdRouterTests/RouterTests.swift
+++ b/Tests/HummingbirdRouterTests/RouterTests.swift
@@ -341,7 +341,7 @@ final class RouterTests: XCTestCase {
 
     func testPartialCapture() async throws {
         let router = HBRouterBuilder(context: HBBasicRouterRequestContext.self) {
-            Get("/files/file.{ext}/{name}.jpg") { _, context -> String in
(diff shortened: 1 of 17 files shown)


Tests/HummingbirdRouterTests/RouterTests.swift
```diff
@@ -341,7 +341,7 @@ final class RouterTests: XCTestCase {
 
     func testPartialCapture() async throws {
         let router = HBRouterBuilder(context: HBBasicRouterRequestContext.self) {
-            Get("/files/file.{ext}/{name}.jpg") { _, context -> String in
+            Route(.get, "/files/file.{ext}/{name}.jpg") { _, context -> String in
```
Joannis, line 344: What's the reason for this change?

Joannis (approved):
I find it extremely difficult to judge when to refer to a symbol the DocC way, with backticks, and when just to name it as text. There are a lot of occurances of types that we could replace with a DocC reference, but I don't know what a good call is there.

PR #350, opened 2024-01-18
Don't require a specific initializer for HBBaseRequestContext

Joannis:
This allows building off HBBaseRequestContext with custom protocols, such as for Lambdas


adam-fowler:
This fixes the issue for lambda. We still have this issue for Channel in the main library. Problem is we have three ways of building a context, one with Channel, one with Lambda event and context, and one with just the allocator and logger. If we want testing we need to support the last one, plus one of the others. I have an idea, using a `HBRequestContextSource` protocol which might resolve this.

I'll merge this for the moment so you can get lambda code in. And I'll look to implement what I'm talking about

adam-fowler (approved):

PR #352, opened 2024-01-18
Remove requirement for makeContext closure

adam-fowler:
If we require `Responder.Context: HBRequestContext` to conform to `HBRequestContext` we can remove the need for a `makeContext` closure


Joannis:
The problem with this is that `HBXCTRouter` becomes unusable with HBLambda

adam-fowler:
Yeah but how do you create an HBXCTRouter if the only `init` requires the context conform `HBRequestContext`

adam-fowler:
@joannis Given we now have HummingbirdLambdaTesting can I resurrect this PR?

PR #353, opened 2024-01-18
Generic request decoder and response encoder

adam-fowler:
Instead of requiring we store a request decoder and response encoder in the core context. This PR makes them accessible from the root of `HBRequestContext`. This allows us to make them to be generic and not existential types.

It does mean we have to get rid of `HBSetCodableMiddleware`, but I think making these generic is a bigger gain.


Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -71,21 +61,21 @@ public struct HBCoreRequestContext: Sendable {
 /// Protocol that all request contexts should conform to. Holds data associated with
 /// a request. Provides context for request processing
 public protocol HBBaseRequestContext: Sendable {
+    associatedtype Decoder: HBRequestDecoder = NullDecoder
+    associatedtype Encoder: HBResponseEncoder = NullEncoder
```
Joannis, line 65: Does it make sense to make these primary associated types?

Joannis, line 65: This will need fixing in HBLambda

Joannis (approved):
One small possible improvement - if it's helpful

Sources/Hummingbird/Server/RequestContext.swift
adam-fowler: I'm not sure that is necessary. When are you going to need to specify the encoder in a parameter type

adam-fowler: yep, PR on its way

PR #354, opened 2024-01-20
2.x.x merge HummingbirdFoundation into Hummingbird

adam-fowler:


Joannis (approved):
LGTM. I feel we should also set the default Codable strategy to JSON now. But that could be a separate PR

PR #355, opened 2024-01-20
Set default encoder/decoder to be JSON

adam-fowler:


Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -99,12 +99,12 @@ extension HBBaseRequestContext {
     public var id: String { self.logger[metadataKey: "hb_id"]!.description }
 }
 
-extension HBBaseRequestContext where Decoder == NullDecoder {
-    public var requestDecoder: Decoder { NullDecoder() }
+extension HBBaseRequestContext where Decoder == JSONDecoder {
+    public var requestDecoder: Decoder { JSONDecoder() }
```
Joannis, line 103: Shouldn't we use ISO8601?

Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -99,12 +99,12 @@ extension HBBaseRequestContext {
...
 }
 
-extension HBBaseRequestContext where Encoder == NullEncoder {
-    public var responseEncoder: Encoder { NullEncoder() }
+extension HBBaseRequestContext where Encoder == JSONEncoder {
+    public var responseEncoder: Encoder { JSONEncoder() }
```
Joannis, line 107: ISO8601?

Sources/Hummingbird/Server/RequestContext.swift
```diff
@@ -99,12 +99,20 @@ extension HBBaseRequestContext {
...
 }
 
-extension HBBaseRequestContext where Decoder == NullDecoder {
-    public var requestDecoder: Decoder { NullDecoder() }
+extension HBBaseRequestContext where Decoder == JSONDecoder {
+    public var requestDecoder: Decoder {
+        let decoder = JSONDecoder()
+        decoder.dateDecodingStrategy = .iso8601
+        return decoder
+    }
```
MahdiBM, line 107: I'd prefer if a `JSONDecoder` / `JSONEncoder` wasn't created every time, but that'll make it a bit complicated to e.g. have a static variable for each and only pass that along every time:
```swift
    private static let requestDecoder: Decoder = {
        let decoder = JSONDecoder()
        decoder.dateDecodingStrategy = .iso8601
        return decoder
    }()
    
    var requestDecoder: Decoder {
        Self.requestDecoder
    }
```
The problem is that users will be able to modify the `JSONDecoder` / `JSONEEncoder` for other requests since they are classes, which could lead to misuse. It'll need some suboptimal hack to make the global decoder/encoder unmodifiable. 🤔

adam-fowler: You can always create your own Context that uses a static copy of a JSONEncoder. But that isn't really recommended. To make JSONEncoder Sendable locks have been added to all the editable fields. Using it across multiple threads will cause stalls as these fields are accessed.

MahdiBM: Valid point. I'm not sure which approach will be faster then ... at which point we can let this be for now.

I have noticed they've marked JSONEncoder as Sendable, but that's just a mistake/bug as I've confirmed with Fabian. JSONEncoder is an open class, and cannot really be Sendable in its current form.

adam-fowler: They made it Sendable by surrounding everything with locks. Ideal they need to make a breaking change and provide a `struct JSONEncoder`.

Joannis (approved):

Issue #356, opened 2024-01-23
HBAsyncMiddleware precondition fail


ronaldmannak:
Even an empty `HBAsyncMiddleware` seems to trigger a precondition failure. Converting the same code to `HBMiddleware` does not crash.

Example code:

```Swift
    struct AsyncMiddleware: HBAsyncMiddleware {
        public func apply(to request: HBRequest, next: HBResponder) async throws -> HBResponse {
            return try await next.respond(to: request)
        }
    }
```

In Configure:

```Swift
 self.middleware.add(AsyncMiddleware())
```

Error: `NIOCore/NIOLoopBound.swift:91: Precondition failed`

adam-fowler:
Do you have a call stack when the precondition is hit?

ronaldmannak:
`OpenAIKeyMiddleware` is the async middleware. When I change `OpenAIKeyMiddleware` to conform `HBMiddleware`, the app runs fine. When I change `OpenAIKeyMiddleware` to conform `HBAsyncMiddleware`, the app crashes.

`OpenAIKeyMiddleware` passes the request on to `ProxyServerMiddleware`, which conforms `HBMiddleware` (not `HBAsyncMiddleware`). It seems that the proxy server is being called.

```
Thread 15 Queue : com.apple.root.default-qos.cooperative (concurrent)
#0	0x000000019dea89d8 in _swift_runtime_on_report ()
#1	0x000000019df669cc in _swift_stdlib_reportFatalErrorInFile ()
#2	0x000000019db55554 in closure #1 in closure #1 in _assertionFailure(_:_:file:line:flags:) ()
#3	0x000000019db54794 in _assertionFailure(_:_:file:line:flags:) ()
#4	0x000000010020a6bc in EventLoop.preconditionInEventLoop(file:line:) at /Users/ronaldmannak/Library/Developer/Xcode/DerivedData/SwiftOpenAIProxy-aauxszkofndrihcsfounfvndgpkv/SourcePackages/checkouts/swift-nio/Sources/NIOCore/EventLoop.swift:1083
#5	0x000000010024d1b8 in NIOLoopBoundBox.__allocating_init(_:eventLoop:) at /Users/ronaldmannak/Library/Developer/Xcode/DerivedData/SwiftOpenAIProxy-aauxszkofndrihcsfounfvndgpkv/SourcePackages/checkouts/swift-nio/Sources/NIOCore/NIOLoopBound.swift:91
[...]

PR #357, opened 2024-01-24
Ensure we are running on EventLoop after HBAsyncMiddleware

adam-fowler:
The next responder was being running in the same task as the async middleware, which could cause problems for routes and middleware after which assume we are still running on the EventLoop

This PR adds flatSubmits to ensure we are running on the correctly EventLoop


Joannis (approved):
Good catch, that's a hard one to spot

adam-fowler:
> Good catch, that's a hard one to spot

Came from #356. So glad to not have to deal with EventLoops anymore

PR #358, opened 2024-01-24
2.x.x remove route options

adam-fowler:
There are currently no options and I don't see there being any in the future. So no point passing around an empty OptionSet


Joannis:
I was thinking of case insensitivity becoming an option, or would we put those on HBRouter instead?

Joannis (approved):

PR #359, opened 2024-01-24
Changes based off comments...

adam-fowler:
From @FranzBusch

- Removed `loggerWithRequestId` global function. It is a one liner and used in two places. This did mean I had to make `RequestID` public but it is hidden behind `@_spi(HBXCT)` for HB internal symbols used in HBXCT.
- `HBApplicationConfiguration` members are all `var`
- Removed `noHTTPServer` as we are not using it
- Removed `threadPoolSize` parameter I missed previously
-  `HBServer. makeServer` is no longer public

The changes to `HBApplicationConfiguration` mean it is now basically just the server configuration parameters so I'll probably merge these two in a later PR.

### DIFF
--- a/Sources/Hummingbird/Server/RequestID.swift
+++ b/Sources/Hummingbird/Server/RequestID.swift
@@ -15,15 +15,14 @@
 import Atomics
 
 /// Generate Unique ID for each request
-@_spi(HBXCT)
-public struct RequestID: CustomStringConvertible {
+package struct RequestID: CustomStringConvertible {
     let low: UInt64
 
-    public init() {
+    package init() {
         self.low = Self.globalRequestID.loadThenWrappingIncrement(by: 1, ordering: .relaxed)
     }
 
-    public var description: String {
+    package var description: String {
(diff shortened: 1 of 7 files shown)


Sources/Hummingbird/Server/RequestID.swift
```diff
@@ -15,14 +15,15 @@
 import Atomics
 
 /// Generate Unique ID for each request
-struct RequestID: CustomStringConvertible {
+@_spi(HBXCT)
+public struct RequestID: CustomStringConvertible {
```
Joannis, line 19: Can we not make this the `package` access modifier?

Joannis (changes requested):

Sources/Hummingbird/Server/RequestID.swift
adam-fowler: I guess we could. I'll change that

adam-fowler: Done

Joannis (approved):

PR #364, opened 2024-01-27
2.x.x - Generalized Client

adam-fowler:
- Added `HBClient` which takes a `HBClientChannel` to setup the channel. 
- `HBClientChannel` differs slightly from `HBChildChannel` in that its handle function can throw errors
- Rename `HBBindAddress` to `HBAddress`
- Cleanup TLS test code, so the getClientTLSConfiguration and getServerTLSConfiguration are in one place
- Added tests for basic HTTP client using `HBClient`.
- Extended TSTLSOptions to support clients as well


Joannis (approved):
LGTM

PR #368, opened 2024-01-31
Make dotenv loading async

adam-fowler:
It isn't actually async at the moment, but will be in the future. So lets set the API correctly


Joannis (approved):
Good call

PR #370, opened 2024-02-02
2.x.x http2 idle state (version 1)

adam-fowler:
I've implemented HTTP2 idle state using the IdleStateHandler and custom HTTP2 stream open and close events. 
The idea being you shutdown when there are no open streams (it is up to the open streams to handle idle state) and an read idle event has been triggered.
The standard HTTP2 events are not generated by the channel handlers created by `configureAsyncHTTPServerPipeline` so I have to create custom events and pass them back to the channel as outbound events to say when streams open and close..

An alternative method can be found in #371


Joannis (approved):

adam-fowler:
closed in favour of #601

PR #371, opened 2024-02-02
2.x.x http2 idle state (version 2)

adam-fowler:
This is an alternative to #370 

Idle state is all done in the async handler. I merge the multiplexer inbound sequence and a async timer sequence together. I iterate over the elements of this merged sequence. For a stream event I start a a handle task for that stream, if a timer event occurs it checks how many streams are open and if the last stream to close was long enough ago. If so then it exits the loop and the channel is closed.

I also merge a gracefulShutdown AsyncStream with the two sequences above to implement graceful shutdown


adam-fowler:
closed in favour of #601

PR #372, opened 2024-02-05
2.x.x - Add support for creating a streamed HBRequestBody

adam-fowler:
One of the problems with moving to wrapping the NIOAsyncChannelInboundStream iterator in the HBRequestBody, is it provided no method for creating a streamed `HBRequestBody` outside of the HTTP request handler. This PR addresses this by internal storing an enum which switches between a `NIOAsyncChannelInboundStream.AsyncIterator` and a `NIOThrowingAsyncSequenceProducer`.

You can now create a new streamed request body as follows
```swift
let (stream, source) = HBRequestBody.makeStream()
```

PR #373, opened 2024-02-05
2.x.x - Use AnyAsyncSequence and creating request body streams

adam-fowler:
This PR is a replacement for #372 

HBRequestBody uses a type erasing `AnyAsyncSequence` to store any AsyncSequence of ByteBuffers
Added support for creating (from Middleware) a NIOThrowingAsyncSequenceProducer to use as a BackPressured async sequence in a request body.

This PR also changes `HBRequestBody` from an `enum` to a `struct` whose contents is hidden.

This PR also allows me to remove `HBRequestBody.collate` and allow users to use `AsyncSequence<ByteBuffer>.collect` from NIO


Joannis (approved):
All of these changes get a big +1 from me

PR #374, opened 2024-02-05
Run processesRunBeforeServerStart after ServiceGroup in router XCT framework

adam-fowler:


Sources/HummingbirdXCT/HBXCTRouter.swift
```diff
@@ -50,10 +50,6 @@ struct HBXCTRouter<Responder: HBResponder>: HBXCTApplication where Responder.Con
             logger: self.logger,
             makeContext: self.makeContext
         )
-        // run the runBeforeServer processes before we start testing
-        for process in self.processesRunBeforeServerStart {
-            try await process()
-        }
         // if we have no services then just run test
         if self.services.count == 0 {
             return try await test(client)
```
Joannis, line 55: We should still run those processes here right?

Joannis (changes requested):

Sources/HummingbirdXCT/HBXCTRouter.swift
adam-fowler: Yep you're right

adam-fowler: fixed

Joannis (approved):

PR #378, opened 2024-02-12
Add primary associated type to HBRouterMethods

sidepelican:
I want to use it like Vapor's `RouteCollection`, so I need the primary associated type to be able to write the following

```swift
struct MyController {
    func boot(routes: some HBRouterMethods<MyRequestContext>) {
        routes.get("foo", use: getFoo)
        routes.post("foo", use: postFoo)
    }
```


adam-fowler (approved):
Fair enough
Of course you can't add middleware, but then you probably shouldn't be allowed to as it would affect routes outside of the controller.

PR #379, opened 2024-02-15
Add request task cancellation, and improve handling of graceful shutdown

Joannis:
Should solve a typical dilemma I've had in systems that get under heavy load:
1. A user connects to download a huge dataset (CSV or XLSX)
2. Server does some heavy queries, and starts outputting CSV/XLSX
3. User gets impatient and refreshes browser, cancelling the request
4. Server is still generating the request that was cancelled
5. Successive attempts by the user become futile and take forever

### DIFF
--- a/Sources/HummingbirdCore/Server/HTTP/HTTPChannelHandler.swift
+++ b/Sources/HummingbirdCore/Server/HTTP/HTTPChannelHandler.swift
@@ -33,83 +33,77 @@ enum HTTPChannelError: Error {
     case closeConnection
 }
 
-enum HTTPState: Int, AtomicValue {
-    case idle
-    case processing
-    case cancelled
-}
-
 extension HTTPChannelHandler {
     public func handleHTTP(asyncChannel: NIOAsyncChannel<HTTPRequestPart, HTTPResponsePart>, logger: Logger) async {
-        let processingRequest = ManagedAtomic(HTTPState.idle)
         do {
-            try await withGracefulShutdownHandler {
-                try await asyncChannel.executeThenClose { inbound, outbound in
-                    let responseWriter = HBHTTPServerBodyWriter(outbound: outbound)
-                    var iterator = inbound.makeAsyncIterator()
+            try await withThrowingDiscardingTaskGroup { taskGroup in
+                taskGroup.addTask {
+                    try await asyncChannel.executeThenClose { inbound, outbound in
+                        let responseWriter = HBHTTPServerBodyWriter(outbound: outbound)
+                        var iterator = inbound.makeAsyncIterator()
 
-                    // read first part, verify it is a head
-                    guard let part = try await iterator.next() else { return }
-                    guard case .head(var head) = part else {
-                        throw HTTPChannelError.unexpectedHTTPPart(part)
-                    }
-
-                    while true {
-                        // set to processing unless it is cancelled then exit
-                        guard processingRequest.exchange(.processing, ordering: .relaxed) == .idle else { break }
-
-                        let bodyStream = NIOAsyncChannelRequestBody(iterator: iterator)
-                        let request = HBRequest(head: head, body: .init(asyncSequence: bodyStream))
-                        let response: HBResponse
-                        do {
-                            response = try await self.responder(request, asyncChannel.channel)
-                        } catch {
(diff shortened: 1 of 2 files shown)


Joannis:
Make sure you hide whitespace on github 
image

Sources/HummingbirdCore/Server/HTTP/HTTPChannelHandler.swift
```diff
@@ -33,83 +33,77 @@ enum HTTPChannelError: Error {
...
-
-                    // read first part, verify it is a head
-                    guard let part = try await iterator.next() else { return }
-                    guard case .head(var head) = part else {
-                        throw HTTPChannelError.unexpectedHTTPPart(part)
-                    }
-
-                    while true {
-                        // set to processing unless it is cancelled then exit
-                        guard processingRequest.exchange(.processing, ordering: .relaxed) == .idle else { break }
```
Joannis, line 59: I didn't add the graceful check here, I wasn't sure if that was "correct" because we already received `.head`

Sources/HummingbirdCore/Server/HTTP/HTTPChannelHandler.swift
```diff
@@ -33,83 +33,77 @@ enum HTTPChannelError: Error {
...
+                                part = try await iterator.next()
+                            }
+
+                            // if part is nil break out of loop
+                            guard let part else {
+                                break
+                            }
+
+                            // Process next request unless it is cancelled then exit
+                            if Task.isShuttingDownGracefully || Task.isCancelled {
```
Joannis, line 89: Cancellation check is moved down. I can imagine that _some_ clients might fail if we do respond, but don't consume the body before closing the connection.

Sources/HummingbirdCore/Server/Server.swift
```diff
@@ -114,7 +114,7 @@ public actor HBServer<ChildChannel: HBChildChannel>: Service {
                         await withDiscardingTaskGroup { group in
                             do {
                                 try await asyncChannel.executeThenClose { inbound in
-                                    for try await childChannel in inbound {
+                                    for try await childChannel in inbound.cancelOnGracefulShutdown() {
```
adam-fowler, line 117: You do need to cancel this on gracefulShutdown, or more requests can come in while finishing the processing of the current request. Again we could possibly use `Task.isShuttingDownGracefully`

Joannis: This change is additive, and adds the `cancelOnGracefulShutdown()`. This wasn't removed, contrary to that it was _missing_.

Sources/HummingbirdCore/Server/HTTP/HTTPChannelHandler.swift
```diff
@@ -33,83 +33,77 @@ enum HTTPChannelError: Error {
...
                     }
                 }
-            } onGracefulShutdown: {
-                // set to cancelled
-                if processingRequest.exchange(.cancelled, ordering: .relaxed) == .idle {
-                    // only close the channel input if it is idle
-                    asyncChannel.channel.close(mode: .input, promise: nil)
-                }
+
+                try await asyncChannel.channel.closeFuture.get()
```
FranzBusch, line 100: This shouldn't be needed. `executeThenClose` makes sure that the channel is closed

Joannis: The reason I do this is because I want to trigger `.cancelAll()` if the channel closes while a request is still being handled

adam-fowler:
In a separate PR can you create a test that demonstrates this is an issue. 

If someone cancels a request would that not close the channel and as soon as the channel is closed writing to it will throw an error and the processing of the request will be cancelled.

adam-fowler:
Or are you looking at the situation where generating the response takes a long time, not writing the response.

Sources/HummingbirdCore/Server/Server.swift
adam-fowler: Channel is closed in `shutdownGracefully` so the cancelOnGracefulShutdown is not needed

adam-fowler (commented):
Removing the graceful shutdown handler means if the connection is open but waiting for http parts it will not shutdown. 

You could probably do this with a middleware where you require the context to have a reference to the channel and then apply it to the routes that might require this kind of support.

PR #385, opened 2024-02-17
2.x.x -CI updates

adam-fowler:
Updated a load of files using https://github.com/hummingbird-project/hummingbird-project-template.
Validate CI is back on so needed to run swiftformat on a few files


.github/workflows/codeql.yml
```diff
@@ -10,6 +10,7 @@ jobs:
   analyze:
     name: Analyze
     runs-on: ubuntu-latest
+    timeout-minutes: 15
```
MahdiBM, line 13: Can't comment there but up in the file you're using `7.x.x` instead of `2.x.x`, i assume due to a copy-pasta error from a project like Soto?

.swiftformat
```diff
@@ -1,14 +1,14 @@
 # Minimum swiftformat version
---minversion 0.47.4
+--minversion 0.51.0
 
 # Swift version
---swiftversion 5.9
+--swiftversion 5.3
```
MahdiBM, line 5: why 5.3 though?

.github/workflows/validate.yml
```diff
@@ -4,18 +4,20 @@ on:
...
+      uses: actions/checkout@v4
       with:
         fetch-depth: 1
     - name: Install Dependencies
       run: |
         brew install mint
-        mint install NickLockwood/SwiftFormat@0.48.17 --no-link
+        mint install NickLockwood/SwiftFormat@0.51.15 --no-link
     - name: run script
       run: ./scripts/validate.sh
```
MahdiBM, line 23: SwiftFormat has a simple `lint` command which can get rid of that script for you.
Not sure about the year validation part of the script though.

.github/workflows/codeql.yml
```diff
@@ -21,7 +22,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v3
+      uses: actions/checkout@v4
```
MahdiBM, line 25: You could have dependabot take care of these.
See https://github.com/vapor/penny-bot/blob/main/.github/dependabot.yml.
You can set the interval to be `weekly` at a specific day and a specific hour too.

.swiftformat
adam-fowler: Missed that

.github/workflows/validate.yml
adam-fowler: Yeah swiftformat can't verify the files have the correct header as far as I can tell. The script is a version of the one Apple uses

.github/workflows/codeql.yml
adam-fowler: Hadn't really been keeping up with this, anyway copied your dependabot setup and removed codeql.yml

.github/workflows/codeql.yml
adam-fowler: Removed

.github/workflows/codeql.yml
MahdiBM: 👍

The Swift part though is unnecessary because this library doesn't contain (or need to contain) any `Package.resolved` files 🙂:
```yml
- package-ecosystem: "swift"
    directory: "/"
    schedule:
      interval: "daily"
    open-pull-requests-limit: 6
    allow:
      - dependency-type: all
    groups:
      all-dependencies:
        patterns:
          - "*"
```

MahdiBM: Also `enable-beta-ecosystems: true` is unnecessary too now, although shouldn't hurt, i thought i've already removed that from Penny's. 😅

Joannis (approved):

PR #386, opened 2024-02-17
Code changes to fix swift-nightly

adam-fowler:
Also fix some of the warnings

### DIFF
--- a/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
+++ b/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
@@ -48,5 +48,10 @@ extension JSONDecoder: HBRequestDecoder {
 /// `HBRequestDecoder` and `HBResponseEncoder` both require conformance to `Sendable`. Given
 /// `JSONEncoder`` and `JSONDecoder`` conform to Sendable in macOS 13+ I think I can just
 /// back date the conformance to all versions of Swift, macOS we support
-extension JSONDecoder: @unchecked Sendable {}
+#if $RetroactiveAttribute
+extension JSONEncoder: @retroactive @unchecked Sendable {}
+extension JSONDecoder: @retroactive @unchecked Sendable {}
+#else
 extension JSONEncoder: @unchecked Sendable {}
+extension JSONDecoder: @unchecked Sendable {}
+#endif
--- a/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
+++ b/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
@@ -48,7 +48,7 @@ extension JSONDecoder: HBRequestDecoder {
 /// `HBRequestDecoder` and `HBResponseEncoder` both require conformance to `Sendable`. Given
 /// `JSONEncoder`` and `JSONDecoder`` conform to Sendable in macOS 13+ I think I can just
 /// back date the conformance to all versions of Swift, macOS we support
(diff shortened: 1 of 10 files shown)


Sources/Hummingbird/Codable/JSON/JSONCoding.swift
```diff
@@ -48,5 +48,10 @@ extension JSONDecoder: HBRequestDecoder {
 /// `HBRequestDecoder` and `HBResponseEncoder` both require conformance to `Sendable`. Given
 /// `JSONEncoder`` and `JSONDecoder`` conform to Sendable in macOS 13+ I think I can just
 /// back date the conformance to all versions of Swift, macOS we support
-extension JSONDecoder: @unchecked Sendable {}
+#if $RetroactiveAttribute
```
MahdiBM, line 51:
```suggestion
#if hasFeature(RetroactiveAttribute)
```
The issue causing `hasFeature` to not work correctly has been resolved on main (not on 5.10, but `RetroactiveAttribute` is not available there anyway.

Sources/Hummingbird/Utils/DateCache+Date.swift
```diff
@@ -17,11 +17,11 @@ import Foundation
 extension HBDateCache {
     /// Date formatter as per section 5.2.14 of RFC 1123
     /// https://www.rfc-editor.org/rfc/rfc1123#page-55
-    static let rfc1123Formatter: DateFormatter = {
+    static var rfc1123Formatter: DateFormatter {
```
MahdiBM, line 20: I assume there were some warnings around this?

You could just use `nonisolated(unsafe)`, considering `DateFormatter`s are known to be expensive, and this is also internal so you know you're not modifying it elsewhere.

Alternatively you could wrap it in another type to avoid mutations, but not sure if that's worth that effort compared to using `nonisolated(unsafe)`.

Tests/HummingbirdTests/URLEncodedForm/URLEncodedNodeTests.swift
```diff
@@ -104,22 +104,27 @@ class URLEncodedFormNodeTests: XCTestCase {
     }
 }
 
-extension URLEncodedFormNode: ExpressibleByStringLiteral {
+#if $RetroactiveAttribute
```
MahdiBM, line 107:
```suggestion
#if hasFeature(RetroactiveAttribute)
```
See the other comment.

Sources/Hummingbird/Utils/DateCache+Date.swift
adam-fowler: Unfortunately we can't really use a global here. DateFormatter on Linux is not thread safe. See https://github.com/apple/swift-corelibs-foundation/issues/3166

Joannis (approved):

Sources/Hummingbird/Utils/DateCache+Date.swift
MahdiBM: I wonder if it is still worth optimizing for.
I don't think what Hummingbird does can trigger that race condition, but even then, a `NIOLockedValueBox` might be better than just making it a computed property.

MahdiBM: Generally speaking, yes, but based on the code, any locking mechanism should be plenty to avoid any crashes in a DateFormatter 🤔

It is suboptimal though, I admit.

PR #391, opened 2024-02-26
Jobs Refactor

adam-fowler:
This refactor is required as it has been pointed out accessing services from a job is quite hard with the current structure. The changes include a breaking up of what defines a job and an actual job instance with parameters. I've also added a JobIdentifier which also includes a reference to the parameter type used by the job to ensure we don't pass bad parameters to a job.

```swift
// define email job parameters
struct EmailJobParameters: Codable {
    let recipient: String
    let subject: String
    let message: String
}

// add identifier for email job and associate with EmailJobParameters type
extension HBJobIdentifier<EmailJobParameters> {
    static var email: Self { .init(name: "email-job") }
}

// Create job queue
let jobQueue = HBJobQueue(.redis(...), maxWorkers: 4, logger: logger)

let ses = SES()

// register email job, with job functionality
jobQueue.registerJob(.email) { parameters, context in
    try await ses.sendEmail(to: parameters.recipient, subject: parameters.subject, body: parameters.message)
}

//  Push email job to queue. Because we have the association between
// the identifier and the required parameters we can ensure the correct parameter type
// is passed in
jobQueue.push(
    id: .email, 
    parameters: .init(recipient: "adam@gmail.com", subject:"Test", message: "Testing, testing, 123")
)
```

### DIFF
--- a/Sources/HummingbirdJobs/JobQueue.swift
+++ b/Sources/HummingbirdJobs/JobQueue.swift
@@ -23,7 +23,8 @@ import ServiceLifecycle
 /// with the queue via either ``HBJobQueue.registerJob(id:maxRetryCount:execute:)`` or
 /// ``HBJobQueue.registerJob(_:)``.
 public struct HBJobQueue<Queue: HBJobQueueDriver>: Service {
-    let queue: Queue
+    /// underlying driver for queue
+    public let queue: Queue
     let handler: HBJobQueueHandler<Queue>
 
     public init(_ queue: Queue, numWorkers: Int = 1, logger: Logger) {
@@ -62,7 +63,7 @@ public struct HBJobQueue<Queue: HBJobQueueDriver>: Service {
     ///  Register job type
     /// - Parameters:
     ///   - job: Job definition
-    public func registerJob<Parameters: Codable & Sendable>(_ job: HBJobDefinition<Parameters>) {
+    public func registerJob(_ job: HBJobDefinition<some Codable & Sendable>) {
         self.handler.registerJob(job)
     }
 
--- a/Sources/HummingbirdJobs/JobQueue.swift
+++ b/Sources/HummingbirdJobs/JobQueue.swift
@@ -26,7 +26,7 @@ public protocol HBJobQueue: AsyncSequence, Sendable where Element == HBQueuedJob
     func onInit() async throws
     /// Push Job onto queue
     /// - Returns: Identifier of queued job
-    @discardableResult func push(_ job: HBJob) async throws -> JobID
+    @discardableResult func push<Parameters: Codable & Sendable>(id: HBJobIdentifier<Parameters>, parameters: Parameters) async throws -> JobID
     /// This is called to say job has finished processing and it can be deleted
     func finished(jobId: JobID) async throws
     /// This is called to say job has failed to run and should be put aside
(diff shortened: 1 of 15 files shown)


adam-fowler:
For reference this is what the Jobs framework did previously with comment indicating the issue I'm trying to fix
```swift
struct EmailJob: HBJob {
    let recipient: String
    let subject: String
    let message: String

    func execute(logger: Logger) async throws {
        // How do I get access to ses to send the email here?
        // I can't include it in the init as the job is recreated by queue
        // at a later date, I can't use variable capture as I would in a closure
    }
}

EmailJob.register()

jobQueue.push(EmailJob(recipient: "adam@gmail.com", subject:"Test", message: "Testing, testing, 123"))
```

Sources/HummingbirdJobs/JobQueueDriver.swift
```diff
@@ -0,0 +1,42 @@
...
+///
+/// Defines how to push and pop job data off a queue
+public protocol HBJobQueueDriver: AsyncSequence, Sendable where Element == HBQueuedJob<JobID> {
+    associatedtype JobID: CustomStringConvertible & Sendable
+
+    /// Called when JobQueueHandler is initialised with this queue
+    func onInit() async throws
+    /// Push Job onto queue
+    /// - Returns: Identifier of queued job
+    func push(data: Data) async throws -> JobID
```
Joannis, line 28: Is there a reason we're using `Data` here, instead of `ByteBuffer`?

Sources/HummingbirdJobs/JobQueue.swift
```diff
@@ -13,31 +13,79 @@
...
-    func finished(jobId: JobID) async throws
-    /// This is called to say job has failed to run and should be put aside
-    func failed(jobId: JobID, error: any Error) async throws
-    /// stop serving jobs
-    func stop() async
-    /// shutdown queue
-    func shutdownGracefully() async
+    @discardableResult public func push<Parameters: Codable & Sendable>(id: HBJobIdentifier<Parameters>, parameters: Parameters) async throws -> Queue.JobID {
+        let jobRequest = HBJobRequest(id: id, parameters: parameters)
+        let data = try JSONEncoder().encode(jobRequest)
```
Joannis, line 42: Would it be possible to allow drivers to encode these parameters (or the request as a whole)? In MongoDB, it would be really unfortunate to encode this to JSON, then add the JSON into MongoDB. Since MongoDB already has (recursive) object support. So it would be able to store this as-is in BSON. Encoding to JSON would prevent indexes from being applies to `id`, and inspecting the data in MongoDB would require quite a lot of work from the developer.

Sources/HummingbirdJobs/JobQueueDriver.swift
adam-fowler: Just the fact we are moving back and forward between the two types. Should probably change to ByteBuffer.

Sources/HummingbirdJobs/JobQueue.swift
adam-fowler: The idea is that the driver only stores data blobs. Do you see an advantage to storing the full object structure? I guess debugging job issues might be easier.

Joannis: Debugging mainly, yes

adam-fowler: I had moved away from the allowing drivers to do the encoding because I was repeating the same code across all of the drivers. It also meant the driver needed to know about the job registry so they could be sure to decode the correct job type. Data blobs meant the driver had a clean split from the job decoding. I'll have another look though.

Joannis (approved):

PR #397, opened 2024-03-11
Require id parameter in registerJob

adam-fowler:


Sources/HummingbirdJobs/JobQueue.swift
```diff
@@ -55,7 +55,7 @@ public struct JobQueue<Queue: JobQueueDriver>: Service {
     ///   - maxRetryCount: Maximum number of times job is retried before being flagged as failed
     ///   - execute: Job code
     public func registerJob<Parameters: Codable & Sendable>(
-        _ id: JobIdentifier<Parameters>,
+        id: HBJobIdentifier<Parameters>,
```
Joannis, line 58: HB prefix is in there?
```suggestion
        id: JobIdentifier<Parameters>,
```

Joannis (changes requested):

Sources/HummingbirdJobs/JobQueue.swift
adam-fowler: shit I shouldn't rush these things

Joannis (approved):

PR #400, opened 2024-03-13
Foundation @preconcurrency only on Linux

adam-fowler:
Clears some warnings from macOS builds


Joannis (approved):
Unfortunate that this is necessary

PR #401, opened 2024-03-16
FileIO Sendable fixes and minor API changes

adam-fowler:
- Don't pass around unsendable FileHandler.
- Use ClosedRanges instead of FileRegions which are also unsendable
- `FileIO.writeFile` can write a generic AsyncSequence instead of just a RequestBody
- Remove usage of EventLoops in FileIO

### DIFF
--- a/Sources/Hummingbird/Files/FileIO.swift
+++ b/Sources/Hummingbird/Files/FileIO.swift
@@ -41,18 +41,8 @@ public struct FileIO: Sendable {
     /// - Returns: Response body
     public func loadFile(path: String, context: some BaseRequestContext) async throws -> ResponseBody {
         do {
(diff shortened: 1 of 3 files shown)


Sources/Hummingbird/Files/FileIO.swift
```diff
@@ -19,15 +19,13 @@ import NIOPosix
 
 /// Manages File reading and writing.
 public struct FileIO: Sendable {
-    let eventLoopGroup: EventLoopGroup
     let fileIO: NonBlockingFileIO
     let chunkSize: Int
 
     /// Initialize FileIO
     /// - Parameter application: application using FileIO
-    public init(eventLoopGroupProvider: EventLoopGroupProvider = .singleton, threadPool: NIOThreadPool = .singleton) {
```
Joannis, line 28: 👍

Joannis (approved):

PR #405, opened 2024-03-19
Use generic for file loading in FileMiddleware

adam-fowler:
This will allow for writing custom file loaders that cache file contents or use another source than the local file system for files

### DIFF
--- a/Sources/Hummingbird/Files/LocalFileSystem.swift
+++ b/Sources/Hummingbird/Files/LocalFileSystem.swift
@@ -1,0 +1,98 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2024 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Foundation
+import Logging
+import NIOPosix
+
+/// Local file system file provider. All file accesses are relative to a root folder
+public struct LocalFileSystem: FileProvider {
+    let rootFolder: String
+    let fileIO: FileIO
+
+    /// Initialize LocalFileSystem FileProvider
+    /// - Parameters:
+    ///   - rootFolder: Root folder to serve files from
+    ///   - threadPool: Thread pool used when loading files
+    ///   - logger: Logger to output root folder information
+    init(rootFolder: String, threadPool: NIOThreadPool, logger: Logger) {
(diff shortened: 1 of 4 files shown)


Sources/Hummingbird/Files/LocalFileSystem.swift
```diff
@@ -0,0 +1,98 @@
...
+        do {
+            let lstat = try await self.fileIO.fileIO.lstat(path: path)
+            let isFolder = (lstat.st_mode & S_IFMT) == S_IFDIR
+            let modificationDate = Double(lstat.st_mtimespec.tv_sec) + (Double(lstat.st_mtimespec.tv_nsec) / 1_000_000_000.0)
+            return .init(
+                isFolder: isFolder,
+                size: numericCast(lstat.st_size),
+                modificationDate: Date(timeIntervalSince1970: modificationDate)
+            )
+        } catch {
```
Joannis, line 75: Is there a known error that would be thrown if the file does not exist?

Joannis (commented):
I like the symbol names a lot, but want to check if the `FileProvider` protocol has all the functions we would need

Sources/Hummingbird/Files/LocalFileSystem.swift
adam-fowler: It's a `NIOCore.IOError(2)` but this could change when we move to NIOFileSystem.

adam-fowler:
> I like the symbol names a lot, but want to check if the `FileProvider` protocol has all the functions we would need

Yeah that is a concern. The FileMiddleware covers pretty much all eventualities. The main possibility would be we need another file attribute.

Sources/Hummingbird/Files/LocalFileSystem.swift
```diff
@@ -0,0 +1,119 @@
...
+    public struct FileAttributes: Sendable, FileMiddlewareFileAttributes {
+        /// Is file a folder
+        public let isFolder: Bool
+        /// Size of file
+        public let size: Int
+        /// Last time file was modified
+        public let modificationDate: Date
+
+        /// Initialize FileAttributes
+        public init(isFolder: Bool, size: Int, modificationDate: Date) {
```
Joannis, line 31: Can this be made `internal` now?

Joannis (changes requested):
Only one change, then it's fine

Sources/Hummingbird/Files/LocalFileSystem.swift
Joannis: I don't think we want to lock ourselves in here

adam-fowler: yeah that should never have been public

adam-fowler: done

Joannis (approved):

PR #406, opened 2024-03-19
Re-instate "hb" prefix on logger, metrics ids

adam-fowler:
Somehow I removed all the "hb" prefixes from the logger and metrics ids. This PR reinstates them


Joannis (approved):
I also didn't see this 😰

PR #409, opened 2024-03-22
Add support for CI that cannot load p12 files

adam-fowler:


Sources/Hummingbird/Server/RequestID.swift
```diff
@@ -15,7 +15,7 @@
 import Atomics
 
 /// Generate Unique ID for each request
-public struct RequestID: CustomStringConvertible {
+public struct RequestID: CustomStringConvertible, Sendable {
```
adam-fowler, line 18: Don't know why but the compiler only just started complaining about this now, but it is a valid complaint

Joannis (approved):

PR #412, opened 2024-03-28
Fix metrics timers

adam-fowler:
Metrics timers were stopping when we returned a response but that is not when the response has been written. The response is written later by the response writer. This moves the recording of metrics to once the response has been written.

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBody.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBody.swift
@@ -78,7 +78,7 @@ public struct ResponseBody: Sendable {
     /// response was written. This functions provides you a method for catching the point when the
     /// response has been fully written. If you drop the response in a middleware run after this
     /// point the post write closure will not get run.
-    public func withPostWriteClosure(_ postWrite: @escaping @Sendable () async -> Void) -> Self {
+    package func withPostWriteClosure(_ postWrite: @escaping @Sendable () async -> Void) -> Self {
         return .init(contentLength: self.contentLength) { writer in
             do {
                 let result = try await self.write(writer)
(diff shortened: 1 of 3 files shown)


Sources/HummingbirdCore/Response/ResponseBody.swift
```diff
@@ -71,6 +71,26 @@ public struct ResponseBody: Sendable {
...
+    /// to the channel
+    ///
+    /// When you return a response from a handler, this cannot be considered to be the point the
+    /// response was written. This functions provides you a method for catching the point when the
+    /// response has been fully written. If you drop the response in a middleware run after this
+    /// point the post write closure will not get run.
+    public func withPostWriteClosure(_ postWrite: @escaping @Sendable () async -> Void) -> Self {
+        return .init(contentLength: self.contentLength) { writer in
+            do {
+                let result = try await self.write(writer)
```
Joannis, line 84: I think it's already the case, but this supports multiple postWrite closures right?

Joannis (approved):

Sources/HummingbirdCore/Response/ResponseBody.swift
```diff
@@ -71,6 +71,26 @@ public struct ResponseBody: Sendable {
...
     }
 
+    /// Create new response body that call a callback once original response body has been written
+    /// to the channel
+    ///
+    /// When you return a response from a handler, this cannot be considered to be the point the
+    /// response was written. This functions provides you a method for catching the point when the
+    /// response has been fully written. If you drop the response in a middleware run after this
+    /// point the post write closure will not get run.
+    public func withPostWriteClosure(_ postWrite: @escaping @Sendable () async -> Void) -> Self {
```
Joannis, line 81: Actually, let's keep this `internal` or at least `package` for now.

Joannis (changes requested):

Sources/HummingbirdCore/Response/ResponseBody.swift
adam-fowler: yes it does

Joannis (approved):

PR #414, opened 2024-04-07
Fix minimum swift-nio version to use async/await FileIO APIs.

sidepelican:
2.0.0-beta.2 uses async/await FileIO APIs in NIO.

Package.swift points to 2.62.0 as the minimum version, but this API was introduced in 2.63.0.
INFO: https://github.com/apple/swift-nio/releases/tag/2.63.0


adam-fowler (approved):

adam-fowler:
Thank you

PR #417, opened 2024-04-18
Minor changes to HTTP channel lhandler

adam-fowler:
- Don't throw connection close error, instead just return from function (it is clearer what is going on)
- Disable outbound header validation. This is already done by swift http types


Joannis (approved):
Great change!

PR #422, opened 2024-04-21
Remove RouteHandler

adam-fowler:
They are awkward to use and it is pretty much impossible to inject dependencies into them


Joannis (approved):
I'm all for removing - just two things:
- Any docs and examples will need to be updated
- What's the expected impact on current users? Do you expect there to be many? If so, maybe we can deprecate first.

adam-fowler:
> I'm all for removing - just two things:
> 
> * Any docs and examples will need to be updated
> * What's the expected impact on current users? Do you expect there to be many? If so, maybe we can deprecate first.

Yeah, we'll need to remove them from the docs
I don't think there will be much impact. They were of some use in 1.0 but given how awkward they are to use in 2.0 I doubt there is much use.

PR #423, opened 2024-04-21
Extend RouterResponder to conform to RouterMiddleware.

adam-fowler:
This means we can add a Router (eg one built with OpenAPI) into a RouterBuilder from HummingbirdRouter


Sources/Hummingbird/Router/RouterResponder.swift
```diff
@@ -57,3 +57,14 @@ public struct RouterResponder<Context: BaseRequestContext>: HTTPResponder {
         return try await responder.respond(to: request, context: context)
     }
 }
+
+extension RouterResponder: RouterMiddleware {
```
Joannis, line 61: I'm not sure I understand the purpose of this change yet. I'm struggling to find any real-world examples that need this.

PR #425, opened 2024-04-23
Update content-length header on setting response body

adam-fowler:
Also add createHeadResponse() to Response so this can be created correctly.


Joannis (approved):
Good find. I thought we already had that

adam-fowler:
> Good find. I thought we already had that

It was Dave Delong who found it. I have always built a Response in one go and never updated `Response.body` after building the `Response` so never found it previously

PR #426, opened 2024-04-23
Remove ServerChildChannel generic associatedtype from ApplicationProtocol

adam-fowler:
Instead use `any ServerChildChannel`. We don’t gain any real benefit from the parameter being a generic over an existential, it gets converted to an existential `Service` anyway. But I’ve had to jump through a number of hoops to ensure we can keep generic parameters where they do make a performance improvement. 

The reason I’m doing this is I was trying to flip between a TLS and non-TLS child channel when build an application and it was really quite awkward as the child channel changed the type of the Application. Removing the ServerChildChannel generic parameter simplifies this considerably.

### DIFF
--- a/Sources/HummingbirdCore/Server/Server.swift
+++ b/Sources/HummingbirdCore/Server/Server.swift
@@ -69,8 +69,11 @@ public actor Server<ChildChannel: ServerChildChannel>: Service {
 
     /// Initialize Server
     /// - Parameters:
-    ///   - group: EventLoopGroup server uses
+    ///   - childChannelSetup: Server child channel
     ///   - configuration: Configuration for server
+    ///   - onServerRunning: Closure to run once server is up and running
+    ///   - eventLoopGroup: EventLoopGroup the server uses
+    ///   = logger: Logger used by server
     public init(
         childChannelSetup: ChildChannel,
         configuration: ServerConfiguration,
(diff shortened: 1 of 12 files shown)


Sources/HummingbirdCore/Server/ServerChildChannel.swift
```diff
@@ -32,3 +33,28 @@ public protocol ServerChildChannel: Sendable {
...
 }
+
+extension ServerChildChannel {
+    /// Build existential ``Server`` from existential `ServerChildChannel`
+    ///
+    /// - Parameters:
+    ///   - configuration: Configuration for server
+    ///   - onServerRunning: Closure to run once server is up and running
+    ///   - eventLoopGroup: EventLoopGroup the server uses
+    ///   = logger: Logger used by server
```
Joannis, line 44:
```suggestion
    ///   - logger: Logger used by server
```

Sources/HummingbirdCore/Server/Server.swift
```diff
@@ -69,8 +69,11 @@ public actor Server<ChildChannel: ServerChildChannel>: Service {
 
     /// Initialize Server
     /// - Parameters:
-    ///   - group: EventLoopGroup server uses
+    ///   - childChannelSetup: Server child channel
     ///   - configuration: Configuration for server
+    ///   - onServerRunning: Closure to run once server is up and running
+    ///   - eventLoopGroup: EventLoopGroup the server uses
+    ///   = logger: Logger used by server
```
Joannis, line 76:
```suggestion
    ///   - logger: Logger used by server
```

Joannis (approved):

PR #430, opened 2024-04-30
Eventloop executor

adam-fowler:
- Add `EventLoopExecutor`
- Add `EventLoopExecutorMap` to map `EventLoop` to an `EventLoopExecutor`
- Require that `ServerChildChannel.Value` conforms to `ChildChannelValue` which returns an EventLoop
- Server then gets the `EventLoop` from the `ServerChildChannel.Value` and runs the Task using the associated `EventLoopExecutor`

### DIFF
--- a/Sources/HummingbirdCore/Server/EventLoopExecutor.swift
+++ b/Sources/HummingbirdCore/Server/EventLoopExecutor.swift
@@ -44,4 +44,20 @@ import NIOCore
             self.eventLoop === other.eventLoop
         }
     }
+
+    struct EventLoopExecutorMap {
+        init(eventLoopGroup: EventLoopGroup) {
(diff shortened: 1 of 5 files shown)


MahdiBM (commented):
`EventLoopExecutor` sounds like something that should exist in NIO? Perhaps worth a PR there?

adam-fowler:
@MahdiBM it probably will in time. They already have a `SerialExecutor`. This PR is experimental at the moment. Unfortunately the results haven't been particularly successful so far. I might be holding it wrong though.

Joannis:
`nonisolated`'s meaning changed, and the custom executor slows things down quite a bit. Should we ditch the idea?

adam-fowler:
We'll close for the moment.

PR #431, opened 2024-04-30
Error messages

adam-fowler:
- Add messages to error responses returned by Hummingbird
- Return messages as json

### DIFF
--- a/Sources/Hummingbird/Router/Parameters+UUID.swift
+++ b/Sources/Hummingbird/Router/Parameters+UUID.swift
@@ -33,7 +33,7 @@ extension Parameters {
         guard let param = self[s[...]],
               let result = UUID(uuidString: String(param))
         else {
-            throw HTTPError(.badRequest)
+            throw HTTPError(.badRequest, message: "Parameter '\(s)' can not be converted to the expected type (UUID)")
         }
         return result
     }
@@ -53,7 +53,7 @@ extension Parameters {
     public func requireAll(_ s: String, as: UUID.Type) throws -> [UUID] {
         return try self[values: s[...]].map {
             guard let result = UUID(uuidString: String($0)) else {
-                throw HTTPError(.badRequest)
+                throw HTTPError(.badRequest, message: "One of the parameters '\(s)' can not be converted to the expected type (UUID)")
             }
             return result
         }
--- a/Sources/Hummingbird/Router/Parameters+UUID.swift
+++ b/Sources/Hummingbird/Router/Parameters+UUID.swift
@@ -30,10 +30,12 @@ extension Parameters {
     ///   - s: parameter id
     ///   - as: type we want returned
     public func require(_ s: String, as: UUID.Type) throws -> UUID {
-        guard let param = self[s[...]],
-              let result = UUID(uuidString: String(param))
+        guard let param = self[s[...]] else {
+            throw HTTPError(.badRequest, message: "Expected parameter does not exist")
+        }
+        guard let result = UUID(uuidString: String(param))
         else {
-            throw HTTPError(.badRequest, message: "Parameter '\(s)' can not be converted to the expected type (UUID)")
+            throw HTTPError(.badRequest, message: "Parameter '\(param)' can not be converted to the expected type (UUID)")
         }
         return result
     }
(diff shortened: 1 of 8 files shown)


Sources/Hummingbird/Files/FileMiddleware.swift
```diff
@@ -98,12 +98,12 @@ public struct FileMiddleware<Context: BaseRequestContext, Provider: FileProvider
 
             // Remove percent encoding from URI path
             guard let path = request.uri.path.removingPercentEncoding else {
-                throw HTTPError(.badRequest)
+                throw HTTPError(.badRequest, message: "Invalid URL")
```
Joannis, line 101:
```suggestion
                throw HTTPError(.badRequest, message: "Invalid percent encoding in URL")
```

Sources/Hummingbird/Files/FileMiddleware.swift
```diff
@@ -98,12 +98,12 @@ public struct FileMiddleware<Context: BaseRequestContext, Provider: FileProvider
...
             }
 
             // file paths that contain ".." are considered illegal
             guard !path.contains("..") else {
-                throw HTTPError(.badRequest)
+                throw HTTPError(.badRequest, message: "URLs including '..' are not supported")
```
Joannis, line 106: I don't think we should provide a message here, this is basically just used maliciously. Any info we give can only help attackers.

Sources/Hummingbird/Router/Parameters+UUID.swift
```diff
@@ -33,7 +33,7 @@ extension Parameters {
         guard let param = self[s[...]],
               let result = UUID(uuidString: String(param))
         else {
-            throw HTTPError(.badRequest)
+            throw HTTPError(.badRequest, message: "Parameter '\(s)' can not be converted to the expected type (UUID)")
```
Joannis, line 36: We shouldn't expose the parameter name here, that's internal information. We can however expose the value provided. I'm also not sure we should provide the expected type per-say.

Sources/Hummingbird/Router/Parameters+UUID.swift
```diff
@@ -53,7 +53,7 @@ extension Parameters {
     public func requireAll(_ s: String, as: UUID.Type) throws -> [UUID] {
         return try self[values: s[...]].map {
             guard let result = UUID(uuidString: String($0)) else {
-                throw HTTPError(.badRequest)
+                throw HTTPError(.badRequest, message: "One of the parameters '\(s)' can not be converted to the expected type (UUID)")
```
Joannis, line 56: Likewise here

Sources/Hummingbird/Router/Parameters.swift
```diff
@@ -46,7 +46,7 @@ public extension Parameters {
     /// - Parameter s: parameter id
     func require(_ s: String) throws -> String {
         guard let param = self[s[...]].map({ String($0) }) else {
-            throw HTTPError(.badRequest)
+            throw HTTPError(.badRequest, message: "Expected parameter '\(s)' does not exist")
```
Joannis, line 49: Likewise here, we should consider not exposing too much info

Joannis (changes requested):
I've got some doubts here, and want to be extra safe. Let's make sure we consider these more thoroughly.

Sources/Hummingbird/Files/FileMiddleware.swift
adam-fowler: removed

Sources/Hummingbird/Router/Parameters+UUID.swift
adam-fowler: Removed parameter name and replaced with text

Sources/Hummingbird/Router/Parameters.swift
adam-fowler: Removed parameter name and replaced with text

adam-fowler:
> Also considering #432, we should consider making empty error messages (possibly optionally) non-json. I don't think it makes sense to serialize `{}` into an `application/json`

Errors with no body no longer report as having a JSON content type

Joannis (approved):

PR #433, opened 2024-05-01
Allow for LogRequestMiddleware to filter the headers it outputs

adam-fowler:
Instead of displaying all the headers, only display ones you care about
```swift
router.middlewares.add(LogRequestMiddleware(.debug, includesHeaders: .some([.contentType, .contentLength]))
```
Options are `.none`, `.all` or `.some([HTTPField.Name])`

### DIFF
--- a/Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
@@ -40,12 +40,12 @@ public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddlewa
         self.includeHeaders = includeHeaders
         // only include headers in the redaction list if we are outputting them
         self.redactHeaders = switch includeHeaders {
-        case .all(let except):
+        case .all(let exceptions):
             // don't include headers in the except list
-            redactHeaders.filter { header in except.first { $0 == header } == nil }
+            redactHeaders.filter { header in !exceptions.contains(header) }
         case .some(let included):
             // only include headers in the included list
-            redactHeaders.filter { header in included.first { $0 == header } != nil }
+            redactHeaders.filter { header in included.contains(header) }
         case .none:
             []
         }
--- a/Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
@@ -18,17 +18,31 @@ import Logging
 /// Middleware outputting to log for every call to server
 public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddleware {
     /// Header filter
-    public enum HeaderFilter: Sendable, ExpressibleByArrayLiteral {
+    public struct HeaderFilter: Sendable, ExpressibleByArrayLiteral {
+        fileprivate enum _Internal: Sendable {
+            case none
+            case all(except: [HTTPField.Name])
+            case some([HTTPField.Name])
+        }
+
+        fileprivate let value: _Internal
+        fileprivate init(_ value: _Internal) {
+            self.value = value
+        }
+
+        /// Don't output any headers
+        public static var none: Self { .init(.none) }
+        /// Output all headers, except the ones indicated
+        public static func all(except: [HTTPField.Name] = []) -> Self { .init(.all(except: except)) }
+        /// Output only these headers
+        public static func some(_ headers: [HTTPField.Name]) -> Self { .init(.some(headers)) }
+
         public typealias ArrayLiteralElement = HTTPField.Name
 
         /// ExpressibleByArrayLiteral requirement
         public init(arrayLiteral elements: ArrayLiteralElement...) {
-            self = .some(elements)
+            self.value = .some(elements)
         }
-
-        case none
-        case all(except: [HTTPField.Name] = [])
-        case some([HTTPField.Name])
     }
 
     let logLevel: Logger.Level
@@ -39,7 +53,7 @@ public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddlewa
         self.logLevel = logLevel
         self.includeHeaders = includeHeaders
         // only include headers in the redaction list if we are outputting them
-        self.redactHeaders = switch includeHeaders {
+        self.redactHeaders = switch includeHeaders.value {
         case .all(let exceptions):
             // don't include headers in the except list
             redactHeaders.filter { header in !exceptions.contains(header) }
@@ -52,7 +66,7 @@ public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddlewa
     }
 
     public func handle(_ request: Request, context: Context, next: (Request, Context) async throws -> Response) async throws -> Response {
-        switch self.includeHeaders {
+        switch self.includeHeaders.value {
         case .none:
             context.logger.log(
                 level: self.logLevel,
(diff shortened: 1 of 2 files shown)


MahdiBM (commented):
Good idea but could be better implemented IMO.
I think we should aim for something like what [traefik](https://doc.traefik.io/traefik/observability/access-logs/) does.
It accepts a `defaultMode`, and you can then provide overrides for specific headers. The setting options are `keep`, `redact`, `drop`.

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
```diff
@@ -12,32 +12,56 @@
 //
 //===----------------------------------------------------------------------===//
 
+import HTTPTypes
 import Logging
 
 /// Middleware outputting to log for every call to server
 public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddleware {
+    public enum HeaderFilter: Sendable {
```
Joannis, line 20: Can we make this `ExpressibleByArrayLiteral` mapped to `.some(..)`?

adam-fowler:
> Good idea but could be better implemented IMO. I think we should aim for something like what [traefik](https://doc.traefik.io/traefik/observability/access-logs/) does. It accepts a `defaultMode`, and you can then provide overrides for specific headers. The setting options are `keep`, `redact`, `drop`.

Given the default mode is none. Not sure I see the point of `drop`. `redact` could be of use though, but we could also add a redaction lists for headers we don't want to report the contents of.

MahdiBM:
@adam-fowler it's useful if you want to log all headers, but not include a few. I've had usages for this in the past, personally.

Generally, how traefik does it is the most flexible way, so users can just do whatever they want, without us making unnecessary assumptions about their needs.

adam-fowler:
> @adam-fowler it's useful if you want to log all headers, but not include a few. I've had usages for this in the past, personally.
> 
> Generally, how traefik does it is the most flexible way, so users can just do whatever they want, without us making unnecessary assumptions about their needs.

So you have two situations I want these specific headers or I want all headers except these ones. Adding an ignore list to `.all` would provide that list for you.

MahdiBM:
Yes, and ignore list would work. But then I'd want to only redact some headers, not ignore them, so I'll need a redact list too

adam-fowler:
> Yes, an ignore list would work. But then I'd want to only redact some headers, not ignore them, so I'll need a redact list too

I'm just looking at the redact list now.

MahdiBM:
Example: show me all headers, redact Authorization or other sensitive headers, and ignore all access origin headers.

adam-fowler:
> Example: show me all headers, redact Authorization or other sensitive headers, and ignore all access control headers.

```swift
let accessControlHeaders: [HTTPFIeld.Name] = [
  . accessControlAllowCredentials,
  . accessControlAllowHeaders,
  . accessControlAllowMethods,
  ...
]
LogRequestsMiddleware(
  .info, 
  includeHeaders: .all(except: accessControlHeaders),
  redact: [.authorization])
}
```

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
```diff
@@ -12,32 +12,111 @@
...
+        case .some(let filter):
             context.logger.log(
                 level: self.logLevel,
-                "",
-                metadata: ["hb_uri": .stringConvertible(request.uri), "hb_method": .string(request.method.rawValue)]
+                "Request",
+                metadata: [
+                    "hb_uri": .stringConvertible(request.uri),
+                    "hb_method": .string(request.method.rawValue),
+                    "hb_headers": .string(self.filterHeaders(headers: request.headers, filter: filter)),
```
MahdiBM, line 82: Not a fan of formatting the headers instead of passing an e.g. Array or Dictionary as `.stringConvertible`.
We use https://github.com/Brainfinance/StackdriverLogging as our swift-log backend which formats stuff properly as a JSON if it can, which makes them look decent on Google Cloud. By passing a String, a logging backend can't try to format the values anymore.

To be clear i haven't seen too many logging backends try to manually format metadata (my own `DiscordLogger` doesn't either), but it does look useful when done appropriately, like with https://github.com/Brainfinance/StackdriverLogging.

MahdiBM (approved):
One concern, otherwise looks good 🙂

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
adam-fowler: Ok makes sense. I can output a dictionary with values from duplicate keys concatenated with a "," inbetween

adam-fowler: done

MahdiBM (approved):

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
```diff
@@ -12,32 +12,103 @@
...
         self.logLevel = logLevel
         self.includeHeaders = includeHeaders
+        // only include headers in the redaction list if we are outputting them
+        self.redactHeaders = switch includeHeaders {
+        case .all(let except):
+            // don't include headers in the except list
+            redactHeaders.filter { header in except.first { $0 == header } == nil }
+        case .some(let included):
+            // only include headers in the included list
+            redactHeaders.filter { header in included.first { $0 == header } != nil }
```
Joannis, line 48:
```suggestion
        case .all(let exceptions):
            // don't include headers in the except list
            redactHeaders.filter { header in !exceptions.contains(header) }
        case .some(let included):
            // only include headers in the included list
            redactHeaders.filter { header in included.contains(header) }
```

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
```diff
@@ -12,32 +12,103 @@
 //
 //===----------------------------------------------------------------------===//
 
+import HTTPTypes
 import Logging
 
 /// Middleware outputting to log for every call to server
 public struct LogRequestsMiddleware<Context: BaseRequestContext>: RouterMiddleware {
+    /// Header filter
+    public enum HeaderFilter: Sendable, ExpressibleByArrayLiteral {
```
Joannis, line 21: What are the odds we'd be adding more cases after this? If the odds are we will add more cases, let's consider making this a `struct` enclosing an `internal enum`

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
```diff
@@ -12,32 +12,103 @@
...
+                    return (key: entry.canonicalName, value: value)
+                }
+            }
+        return .init(headers) { "\($0), \($1)" }
+    }
+
+    func allHeaders(headers: HTTPFields, except: [HTTPField.Name]) -> [String: String] {
+        let headers = headers
+            .compactMap { entry -> (key: String, value: String)? in
+                guard except.first(where: { entry.name == $0 }) == nil else { return nil }
```
Joannis, line 105:
```suggestion
                if except.contains(where: { entry.name == $0 }) { return nil }
```

Joannis: Note that I converted to `if`, since `guard !` is a double-not, so not great.

Joannis (commented):
Looks good.

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
adam-fowler: I prefer `except`. It reads better. All headers except. But I'll change to contains

Sources/Hummingbird/Middleware/LogRequestMiddleware.swift
adam-fowler: Yeah its always worthwhile

Joannis (approved):

PR #435, opened 2024-05-02
Report correct length for trace span

adam-fowler:
The trace span in the `TracingMiddleware` should be the length it takes to return a response. Unfortunately because our Response body is a closure that will write the body at a later point and not the actual body we have not written the response body to the channel by the time we end the span in the `TracingMiddleware`. 

To fix this I am manually starting and ending the span myself instead of using `withSpan`. The `TracingMiddleware` returns a response body that ends the span once the body has been written.

Another issue arises if I leave this as is though. Someone could drop the response body which was returned by the `TracingMiddleware` and the `span.end` would never be called. To get around this I have wrapped it in a small class which will call `span.end()` on `deinit`, if it hasn't already been called.


Joannis (approved):
'been trying to think of a better solution - don't think there is one.

PR #441, opened 2024-05-08
Rely on NIO defaults for child maxMessagePerRead

adam-fowler:
One message per read is probably too low. Let's revert to the NIO default of 4


Joannis:
Is there a noticable performance improvement?

PR #444, opened 2024-05-08
Allow specialization of the RouterResponder

Joannis:
I've ran the `PerformanceTest` codebase two times, before and after this PR. Note that the PerformanceTest only uses 4 cores for HB2 - so that we have resource budget for regular OS use and the `wrk` client benckmark suite.

Since I made a lot of symbols inlinable, I had to make them public or internal. Public with SPI does not work with inlinable. We'll need to decide on a public API surface if we want benchmarks.

TODOs:
- [ ] Fix the router benchmarks in the Benchmarks folder.

## Before

```
Running 15s test @ http://localhost:8080
  8 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.78ms    1.35ms  47.83ms   91.64%
    Req/Sec     9.80k     1.47k   26.85k    89.11%
  1173503 requests in 15.10s, 166.75MB read
Requests/sec:  77697.81
Transfer/sec:     11.04MB
```

## After

```
Running 15s test @ http://localhost:8080
  8 threads and 128 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     1.69ms    1.37ms  50.63ms   93.09%
    Req/Sec    10.28k   694.90    13.73k    76.45%
  1233179 requests in 15.10s, 175.23MB read
Requests/sec:  81659.66
Transfer/sec:     11.60MB
```

### DIFF
--- a/Sources/Hummingbird/Router/TrieRouter.swift
+++ b/Sources/Hummingbird/Router/TrieRouter.swift
@@ -15,10 +15,12 @@
 import HummingbirdCore
 
 /// URI Path Trie Builder
-@_spi(Internal) public struct RouterPathTrieBuilder<Value: Sendable> {
+@_documentation(visibility: internal)
+public struct RouterPathTrieBuilder<Value: Sendable> {
     @usableFromInline
     var root: Node
 
+    @_documentation(visibility: internal)
     public init() {
         self.root = Node(key: .null, output: nil)
     }
@@ -50,7 +52,8 @@ import HummingbirdCore
     }
 
     /// Trie Node. Each node represents one component of a URI path
-    @_spi(Internal) public final class Node {
+    @_documentation(visibility: internal)
+    public final class Node {
         @usableFromInline
         let key: RouterPath.Element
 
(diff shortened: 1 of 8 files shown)


adam-fowler:
Can you get the Benchmarks and Tests working again, to indicate what we would need to make public. I guess for the tests it doesn't matter too much as you can import as `@testable`.

adam-fowler (commented):
One comment above

Sources/Hummingbird/Router/TrieRouter.swift
```diff
@@ -15,10 +15,12 @@
 import HummingbirdCore
 
 /// URI Path Trie Builder
-@_spi(Internal) public struct RouterPathTrieBuilder<Value: Sendable> {
+@_documentation(visibility: internal)
+public struct RouterPathTrieBuilder<Value: Sendable> {
```
adam-fowler, line 19: This shouldn't need to be public. It is only ever used to build a router. I'm not sure why the TrieRouter serialize functions have been made `@inlineable`. They don't need to be. Building a router doesn't need to be optmized., It is only the resolve functions that should be `@inlineable`. By removing the inalienable on the serialize functions this can be make internal again.

adam-fowler (commented):
One issue, that was inherited from the original array of structs PR

adam-fowler (approved):
One minor compile error I fixed, but otherwise looks good

PR #448, opened 2024-05-13
Fix up some old docs and add more DocC annotations

Joannis:
A lot of the DocC symbol links don't work in DocC, but they don't hurt and it helps the rendering on swiftinit.org and swiftonserver.com


adam-fowler:
Adding links to outside libraries produces errors while building the docs, which then hide real errors in the documentation.The documentation pipeline currently requires that docs don't generate errors before publishing. I'd rather not include links to outside libraries for this reason

PR #450, opened 2024-05-14
Add requirement that channel is accessible in ServerChildChannel.Value

adam-fowler:
This adds an additional requirement on the Value passed around by a ServerChildChannel that the Channel that spawned the child channel is accessible from the Value. 

This PR doesn't add any additional functionality. It is here to prepare for the situation where we need the EventLoop for the Task executor.

### DIFF
--- a/Sources/HummingbirdCore/Server/ServerChildChannel.swift
+++ b/Sources/HummingbirdCore/Server/ServerChildChannel.swift
@@ -16,9 +16,15 @@ import Logging
 import NIOCore
 import ServiceLifecycle
 
-/// HTTPServer child channel setup protocol
+/// Protocol for typed server child channel
+public protocol ServerChildChannelValue: Sendable {
+    /// Child channel that spawned child channel
+    var channel: Channel { get }
+}
+
+/// Generic server child channel setup protocol
 public protocol ServerChildChannel: Sendable {
-    associatedtype Value: Sendable
+    associatedtype Value: ServerChildChannelValue
 
     /// Setup child channel
     /// - Parameters:
--- a/Sources/HummingbirdHTTP2/HTTP2Channel.swift
+++ b/Sources/HummingbirdHTTP2/HTTP2Channel.swift
@@ -26,7 +26,7 @@ import NIOSSL
 /// Child channel for processing HTTP1 with the option of upgrading to HTTP2
 public struct HTTP2UpgradeChannel: HTTPChannelHandler {
     public struct Value: ServerChildChannelValue {
-        let negociatedHTTPVersion: EventLoopFuture<NIONegotiatedHTTPVersion<HTTP1Channel.Value, (NIOAsyncChannel<HTTP2Frame, HTTP2Frame>, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP1Channel.Value>)>>
+        let negotiatedHTTPVersion: EventLoopFuture<NIONegotiatedHTTPVersion<HTTP1Channel.Value, (NIOAsyncChannel<HTTP2Frame, HTTP2Frame>, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP1Channel.Value>)>>
         public let channel: Channel
     }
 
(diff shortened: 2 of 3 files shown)


Sources/HummingbirdCore/Server/ServerChildChannel.swift
```diff
@@ -16,9 +16,15 @@ import Logging
 import NIOCore
 import ServiceLifecycle
 
-/// HTTPServer child channel setup protocol
+/// Protocol for typed server child channel
+public protocol ServerChildChannelValue: Sendable {
+    /// Child channel that spawned child channel
+    var channel: Channel { get }
```
Joannis, line 22: Isn't it better if we add `var eventLoop: EventLoop` here? This assumes we're using a `NIO.Channel`, whereas we could have other implementations such as Lambda that aren't connected to a socket at all.

I would even consider removing this requirement entirely, and replacing it with something like this:

```swift
func addTask<Failure: Error>(
  _ operation: @escaping () async throws -> Void,
  to taskGroup: inout ThrowingDiscardingTaskGroup<Failure>
)
```

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -25,7 +25,10 @@ import NIOSSL
 
 /// Child channel for processing HTTP1 with the option of upgrading to HTTP2
 public struct HTTP2UpgradeChannel: HTTPChannelHandler {
-    public typealias Value = EventLoopFuture<NIONegotiatedHTTPVersion<HTTP1Channel.Value, (NIOAsyncChannel<HTTP2Frame, HTTP2Frame>, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP1Channel.Value>)>>
+    public struct Value: ServerChildChannelValue {
+        let negociatedHTTPVersion: EventLoopFuture<NIONegotiatedHTTPVersion<HTTP1Channel.Value, (NIOAsyncChannel<HTTP2Frame, HTTP2Frame>, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP1Channel.Value>)>>
```
Joannis, line 29:
```suggestion
        let negotiatedHTTPVersion: EventLoopFuture<NIONegotiatedHTTPVersion<HTTP1Channel.Value, (NIOAsyncChannel<HTTP2Frame, HTTP2Frame>, NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP1Channel.Value>)>>
```

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -91,6 +94,8 @@ public struct HTTP2UpgradeChannel: HTTPChannelHandler {
                 }.flatMapThrowing {
                     try HTTP1Channel.Value(wrappingChannelSynchronously: http2ChildChannel)
                 }
+        }.map {
+            .init(negociatedHTTPVersion: $0, channel: channel)
```
Joannis, line 98:
```suggestion
            .init(negotiatedHTTPVersion: $0, channel: channel)
```

Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -100,7 +105,7 @@ public struct HTTP2UpgradeChannel: HTTPChannelHandler {
     ///   - logger: Logger to use while processing messages
     public func handle(value: Value, logger: Logger) async {
         do {
-            let channel = try await value.get()
+            let channel = try await value.negociatedHTTPVersion.get()
```
Joannis, line 108:
```suggestion
            let channel = try await value.negotiatedHTTPVersion.get()
```

Joannis (changes requested):

Sources/HummingbirdCore/Server/ServerChildChannel.swift
adam-fowler: The `ServerChildChannel` assumes we are using a NIO.Channel anyway. It is there to setup a NIO child channel. This code is never run on lambda.

I'm not exactly sure how you would use the `addTask` function

Joannis (approved):

PR #452, opened 2024-05-16
API Cleanup: HTTPChannelBuilder should build server not channel

adam-fowler:
Also renamed it to HTTPServerBuilder

Also rename Address back to BindAddress now that it is only being used as a bind address (It got named to Address when  I introduced the generic client, which has been removed now)

### DIFF
--- a/Sources/HummingbirdCore/Server/HTTP/HTTPServerBuilder.swift
+++ b/Sources/HummingbirdCore/Server/HTTP/HTTPServerBuilder.swift
@@ -43,7 +43,7 @@ public struct HTTPServerBuilder: Sendable {
         eventLoopGroup: EventLoopGroup,
         logger: Logger,
         responder: @escaping HTTPChannelHandler.Responder,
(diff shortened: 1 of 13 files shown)


Sources/HummingbirdCore/Server/HTTP/HTTPServerBuilder.swift
```diff
@@ -12,25 +12,46 @@
...
+    ///   - logger: Logger used by server
+    ///   - responder: HTTP responder
+    ///   - onServerRunning: Closure to run once server is up and running
+    /// - Returns: Server Service
+    public func buildServer(
+        configuration: ServerConfiguration,
+        eventLoopGroup: EventLoopGroup,
+        logger: Logger,
+        responder: @escaping HTTPChannelHandler.Responder,
+        onServerRunning: (@Sendable (Channel) async -> Void)? = { _ in }
```
Joannis, line 46: Do we use an empty closure as a default instead of `nil` for a reason?

Joannis (approved):
I think that's a nice improvement

PR #454, opened 2024-05-19
Minor optimisations

adam-fowler:
Don't allocate ByteBuffer before we know its size.
Provide contentLength when generating responses so header field size is not recalculated, when contentLength is added
Use `HTTPFields.contains` instead of subscript which constructs contents of HTTPField

### DIFF
--- a/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
+++ b/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
@@ -24,8 +24,7 @@ extension JSONEncoder: ResponseEncoder {
     ///   - request: Request used to generate response
     public func encode(_ value: some Encodable, from request: Request, context: some BaseRequestContext) throws -> Response {
         let data = try self.encode(value)
-        var buffer = context.allocator.buffer(capacity: data.count)
-        buffer.writeBytes(data)
+        let buffer = context.allocator.buffer(data: data)
         return Response(
             status: .ok,
             headers: [
--- a/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
+++ b/Sources/Hummingbird/Codable/JSON/JSONCoding.swift
@@ -23,12 +23,15 @@ extension JSONEncoder: ResponseEncoder {
     ///   - value: Value to encode
     ///   - request: Request used to generate response
     public func encode(_ value: some Encodable, from request: Request, context: some BaseRequestContext) throws -> Response {
-        var buffer = context.allocator.buffer(capacity: 0)
         let data = try self.encode(value)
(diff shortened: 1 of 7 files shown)


Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedForm+Request.swift
```diff
@@ -18,12 +18,15 @@ extension URLEncodedFormEncoder: ResponseEncoder {
     ///   - value: Value to encode
     ///   - request: Request used to generate response
     public func encode(_ value: some Encodable, from request: Request, context: some BaseRequestContext) throws -> Response {
-        var buffer = context.allocator.buffer(capacity: 0)
         let string = try self.encode(value)
+        var buffer = context.allocator.buffer(capacity: string.utf8.count)
```
Joannis, line 22: Why not create the buffer from the string using NIO's helpers?

Sources/Hummingbird/Codable/JSON/JSONCoding.swift
```diff
@@ -23,12 +23,15 @@ extension JSONEncoder: ResponseEncoder {
     ///   - value: Value to encode
     ///   - request: Request used to generate response
     public func encode(_ value: some Encodable, from request: Request, context: some BaseRequestContext) throws -> Response {
-        var buffer = context.allocator.buffer(capacity: 0)
         let data = try self.encode(value)
+        var buffer = context.allocator.buffer(capacity: data.count)
```
Joannis, line 27: Can we use NIO's allocator helpers here

Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedForm+Request.swift
adam-fowler: No because that uses a temporary allocator and not the one that was attached to the Channel

Sources/Hummingbird/Codable/JSON/JSONCoding.swift
Joannis: [For bytes](https://github.com/apple/swift-nio/blob/609469f9689603c8419f1402dd984b0faa763c54/Sources/NIOCore/ByteBuffer-aux.swift#L739-L742)

Joannis (approved):
I don't mind merging this as-is. But it seems sensible to me to merge the call to allocate and write into one.

Sources/Hummingbird/Codable/JSON/JSONCoding.swift
adam-fowler: Fair point. Will fix

PR #455, opened 2024-05-19
Remove NIOLockedValueBox from EndpointPath

Joannis:
It's not needed (anymore) because we expect the entire context setup to be value types

### DIFF
--- a/Sources/Hummingbird/Server/RequestContext.swift
+++ b/Sources/Hummingbird/Server/RequestContext.swift
@@ -21,17 +21,10 @@ import Tracing
 
 /// Endpoint path storage
 public struct EndpointPath: Sendable {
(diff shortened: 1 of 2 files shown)


adam-fowler:
You should run tests before, committing stuff. This is needed at the moment as the router path is resolved after middleware are run in the HummingbirdRouter. So the endpoint needs to be a reference. An alternative optimisation would be to add a protocol 

```swift
protocol EndpointRequestContext {
    let endpoint: Endpoint
}
``` 
and require that the tracing and metrics middleware are only available when the request context conforms to this protocol. They are the only times the endpoint is needed

Joannis:
Ah I forgot about HummingbirdRouter, sorry!

PR #456, opened 2024-05-19
Optimise FlatDictionary init for scenarios where no values are stored and llow inlining access

Joannis:
This provides specialized access to the type, and virtually eliminates its presence from routes where a flat dictionary is not used. This affects the route parameters primarily, but helps in particular because `CoreRequestContext` _always_ starts out with an empty `Parameters` instance.

### DIFF
--- a/Sources/HummingbirdCore/Utils/FlatDictionary.swift
+++ b/Sources/HummingbirdCore/Utils/FlatDictionary.swift
@@ -175,4 +175,5 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
 }
 
 // FlatDictionary is Sendable when Key and Value are Sendable
+extension FlatDictionary.Storage: Sendable where Key: Sendable, Value: Sendable {}
 extension FlatDictionary: Sendable where Key: Sendable, Value: Sendable {}
--- a/Sources/HummingbirdCore/Utils/FlatDictionary.swift
+++ b/Sources/HummingbirdCore/Utils/FlatDictionary.swift
@@ -26,76 +26,112 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
     public typealias Element = (key: Key, value: Value)
     public typealias Index = Array<Element>.Index
 
+    @usableFromInline
+    struct Storage {
+        @usableFromInline internal /* private */ var elements: [Element]
+        @usableFromInline internal /* private */ var hashKeys: [Int]
+
+        @usableFromInline
+        init(elements: [Element], hashKeys: [Int]) {
+            self.elements = elements
+            self.hashKeys = hashKeys
+        }
+    }
+
+    @usableFromInline internal /* private */ var storage: Storage?
+
     // MARK: Collection requirements
 
     /// The position of the first element
-    public var startIndex: Index { self.elements.startIndex }
+    @inlinable
+    public var startIndex: Index { self.storage?.elements.startIndex ?? 0 }
     /// The position of the element just after the last element
-    public var endIndex: Index { self.elements.endIndex }
+    @inlinable
+    public var endIndex: Index { self.storage?.elements.endIndex ?? 0 }
     /// Access element at specific position
-    public subscript(_ index: Index) -> Element { return self.elements[index] }
+    @inlinable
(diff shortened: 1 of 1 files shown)


adam-fowler:
Empty arrays don't allocate any memory, so I'm not sure this is going to be a saving, as you have added a branch whenever accessing the parameters now

Joannis:
I think my measurements are macOS-specific then.

Sources/HummingbirdCore/Utils/FlatDictionary.swift
```diff
@@ -26,104 +26,154 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
     public typealias Element = (key: Key, value: Value)
     public typealias Index = Array<Element>.Index
 
+    @usableFromInline
+    struct Storage {
+        @usableFromInline internal /* private */ var elements: [Element]
```
adam-fowler, line 31: Can we remove the commented out `/* private */` if we require everything to be `internal`

Sources/HummingbirdCore/Utils/FlatDictionary.swift
```diff
@@ -26,104 +26,154 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
...
+    @inlinable
+    public var startIndex: Index { self.storage?.elements.startIndex ?? 0 }
     /// The position of the element just after the last element
-    public var endIndex: Index { self.elements.endIndex }
+    @inlinable
+    public var endIndex: Index { self.storage?.elements.endIndex ?? 0 }
     /// Access element at specific position
-    public subscript(_ index: Index) -> Element { return self.elements[index] }
+    @inlinable
+    public subscript(_ index: Index) -> Element { return self.storage!.elements[index] }
```
adam-fowler, line 53: `!` isn't ideal here, perhaps we should add a precondition that `storage` exist and provide an out of range message or something similar, given the index will be out of range

Sources/HummingbirdCore/Utils/FlatDictionary.swift
```diff
@@ -26,104 +26,154 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
...
     /// Returns the index immediately after the given index
-    public func index(after index: Index) -> Index { self.elements.index(after: index) }
+    @inlinable
+    public func index(after index: Index) -> Index { index + 1 }
```
adam-fowler, line 56: Any reason you changed this? I know `index+1` is correct in most cases, but what is the expected result when `index == endIndex`?

Sources/HummingbirdCore/Utils/FlatDictionary.swift
```diff
@@ -26,104 +26,154 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
...
+            if let storage, let index = storage.hashKeys.firstIndex(of: hashKey) {
+                return storage.elements[index].value
             } else {
                 return nil
             }
         }
         set {
             let hashKey = Self.hashKey(key)
-            if let index = hashKeys.firstIndex(of: hashKey) {
+            if var storage, let index = storage.hashKeys.firstIndex(of: hashKey) {
```
adam-fowler, line 98: The ordering of the if statements here is slightly confusing. Currently you test for the existence of storage twice when the key doesn't exist. Perhaps it should be 
```swift
if let storage {
    // do stuff when storage already exists
} else {
    // create storage if needed
}
```

Sources/HummingbirdCore/Utils/FlatDictionary.swift
```diff
@@ -26,104 +26,154 @@ public struct FlatDictionary<Key: Hashable, Value>: Collection, ExpressibleByDic
...
         hasher.combine(key)
         return hasher.finalize()
     }
-
-    private var elements: [Element]
-    private var hashKeys: [Int]
 }
 
 // FlatDictionary is Sendable when Key and Value are Sendable
+extension FlatDictionary.Storage: Sendable where Key: Sendable, Value: Sendable {}
```
adam-fowler, line 178: I added this in to get rid of a Sendable warning

adam-fowler (changes requested):
Looks good, some minor change requests

adam-fowler:
@Joannis can I close this?

PR #457, opened 2024-05-19
Provide content-length in the HTTPFields' literal when creating a response

Joannis:
This allows the HTTPFields to allocate sufficient room for `Content-Length`, `Server` and `Date`, preventing reallocs

### DIFF
--- a/Sources/Hummingbird/Utils/HTTPFields+Payload.swift
+++ b/Sources/Hummingbird/Utils/HTTPFields+Payload.swift
@@ -1,0 +1,14 @@
+import HTTPTypes
+
+extension HTTPFields {
+    init(contentType: String, contentLength: Int) {
+        self.init()
+
+        // Content-Type, Content-Length, Server, Date + 2 extra headers
+        // This should cover our expected amount of headers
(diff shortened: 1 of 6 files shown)


Joannis:
The main issue was that contentl-length wasn't set yet, so the HTTPFields had a size of `1` due to `reserveCapacity` being called with `1` (from the literal). Then we were _immediately_ adding our content-length on top of the literal, finishing it off with `server` and `date` headers.

Sources/Hummingbird/Utils/HTTPFields+Payload.swift
```diff
@@ -0,0 +1,14 @@
+import HTTPTypes
+
+extension HTTPFields {
+    init(contentType: String, contentLength: Int) {
```
Joannis, line 4: I wanted to centralize this as an internal helper, so that we can tweak the formula from here

Joannis:
This is likely a great example of something that could use optimisation in HTTPFields

adam-fowler:
This was already covered in PR#454

PR #458, opened 2024-05-19
Prevent unnecessary Copy-on-Write behaviour in HTTPFields

Joannis:


Sources/HummingbirdCore/Response/Response.swift
```diff
@@ -16,32 +16,34 @@ import HTTPTypes
...
-        get { self.head.status }
-        set { self.head.status = newValue }
-    }
-
-    public var headers: HTTPFields {
-        get { self.head.headerFields }
-        set { self.head.headerFields = newValue }
-    }
+    public var status: HTTPResponse.Status
+    public var headers: HTTPFields
```
Joannis, line 46: First of all, modifying `response.headers` was leading to a CoW. This is because `headers` was a getter/setter, thus copied the value from `head`. I've reversed this relationship

Sources/HummingbirdCore/Response/Response.swift
```diff
@@ -16,32 +16,34 @@ import HTTPTypes
...
     }
 
     public init(status: HTTPResponse.Status, headers: HTTPFields = .init(), body: ResponseBody = .init()) {
-        self.head = .init(status: status, headerFields: headers)
+        self.status = status
+        self.headers = headers
         self.body = body
-        if let contentLength = body.contentLength, headers[.contentLength] == nil {
-            self.head.headerFields[.contentLength] = String(describing: contentLength)
+        if let contentLength = body.contentLength, self.headers[.contentLength] == nil {
```
Joannis, line 40: `self.` was missing, leading to the original `headers` still being references. This lead to there being two active copies, which would cause a CoW

adam-fowler:
These changes were merged with #454 so this can be closed

PR #462, opened 2024-05-23
Make HTTP handler non-throwing. Application should do error handling

adam-fowler:


Sources/Hummingbird/Application.swift
```diff
@@ -115,6 +115,7 @@ extension ApplicationProtocol {
                     response = httpError.response(allocator: channel.allocator)
                 default:
                     // this error has not been recognised
+                    context.logger.debug("Error: \(error)")
```
Joannis, line 118: Given recent discussions on slack between Gwynne and Johannes, I think we should consider adding the error as metadata, clarifying the debug message, and make the main message a constant string.

Joannis (approved):

Sources/Hummingbird/Application.swift
adam-fowler: My bad. I meant to do that. I remember a long conversation when swift-log was first being setup which said pretty much the same thing

PR #463, opened 2024-05-25
Add initializer for HTTPFields that reserves space for standard headers

adam-fowler:


Sources/Hummingbird/Router/ResponseGenerator.swift
```diff
@@ -128,3 +128,20 @@ public struct EditedResponse<Generator: ResponseGenerator>: ResponseGenerator {
...
     }
 }
+
+extension HTTPFields {
+    /// Initialize HTTPFields with contentType and contentLength headers and also reserve
+    /// space for server and date headers which will be set later
+    /// - Parameters:
+    ///   - contentType: Content Type header
+    ///   - contentLength: Content Length
+    init(
```
MahdiBM, line 138: I'd prefer a static func like `static func withExtraCapacity` for better readabilty/clarity, even though this is only an internal function, unless it makes some kind of non-insignificant difference in the compiled code (I know `init`s have some special properties, but i've never cared enough for such optimizations to actually go after it)

adam-fowler: Fair enough wasn't totally happy with the `init` I must say.

Joannis (approved):

PR #464, opened 2024-05-27
Fix router path precedence, parsing of parameter paths

adam-fowler:
parameter with prefix/suffix should not have precedence over a fully specified path name 
Also fixed parsing of path elements so `{param}` is parsed as `.parameter` and not `.prefixParameter`

### DIFF
--- a/Tests/HummingbirdTests/TrieRouterTests.swift
+++ b/Tests/HummingbirdTests/TrieRouterTests.swift
@@ -77,11 +77,15 @@ class TrieRouterTests: XCTestCase {
         let trieBuilder = RouterPathTrieBuilder<String>()
         trieBuilder.addEntry("users/:user", value: "test1")
         trieBuilder.addEntry("users/:user/name", value: "john smith")
+        trieBuilder.addEntry("users/:user/name/{id}", value: "41D2DF67-C2C2-4842-B1DA-9F4549BED3F0")
         let trie = trieBuilder.build()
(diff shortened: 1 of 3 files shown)


adam-fowler:
@tib this should fix your issue

Tests/HummingbirdTests/TrieRouterTests.swift
```diff
@@ -29,6 +29,30 @@ class TrieRouterTests: XCTestCase {
         XCTAssertEqual(trie.resolve("/Users/jane/bin")?.value, "test3")
     }
 
+    func testPathParsing() {
+        let trieBuilder = RouterPathTrieBuilder<String>()
+        func getFirstChildElement(_ path: String) -> RouterPath.Element? {
+            trieBuilder.root.children.first(where: { $0.key == path })?.children.first?.key
+        }
+        trieBuilder.addEntry("test1/:param", value: "*")
```
tib, line 37: Just a quick question: shouldn't we have a test for something like this:

```swift
"foo/{fooId}"
"foo/{fooId}/bar/{barId}
```

adam-fowler: This test is just there to check the parsing of individual path components. I've added another test in `testGetParameters` which does pretty much what you are asking

tib (approved):

Joannis (approved):

PR #468, opened 2024-06-10
Rename collateBody to collectBody

adam-fowler:
Also only require the maxUploadSize instead of the full context


adam-fowler:
@sliemeobn

Joannis (approved):
I think this is a nice change, but can we add the old symbol back with deprecation until release?

adam-fowler:
> I think this is a nice change, but can we add the old symbol back with deprecation until release?

Parameters have changed so that won't work. I guess I could add a wrapper function though

adam-fowler:
@Joannis added deprecated method, plus I moved the new function to Core as it doesn't use any Hummingbird symbols

PR #469, opened 2024-06-10
Replace RequestContext initialization parameters with associatedtype

adam-fowler:
Instead of 
```swift
struct MyRequestContext: RequestContext {
    var coreContext: CoreRequestContext
    init(channel: Channel, logger: Logger) {
        self.coreContext = .init(allocator: channel.allocator, logger: logger)
    }
}
```
You have where, `Source` is the source of the request
```swift
struct MyRequestContext: RequestContext {
    var coreContext: CoreRequestContext
    init(source: Source) {
        self.coreContext = .init(source: Source)
    }
}
```
In the server situation `Source` is a struct that includes the logger and the channel. You can still access the channel via `source.channel`. In the lambda situation this would include the Event that triggered the lambda and the LambdaContext.

It makes context creation simpler and hides implementation details

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -41,7 +41,8 @@ public enum EventLoopGroupProvider {
     }
 }
 
-public protocol ApplicationProtocol: Service where Context: RequestContext {
+/// Protocol for an Application. Brings together all the components of Hummingbird together
+public protocol ApplicationProtocol: Service where Context: InstantiableRequestContext, Context.Source == ServerRequestContextSource {
     /// Responder that generates a response from a requests and context
     associatedtype Responder: HTTPResponder
     /// Context passed with Request to responder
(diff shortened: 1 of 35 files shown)


Joannis (commented):
I like this a lot. My only "concern" is that we lose the ability to add more properties to `CoreRequestContext` later with this setup. With the old setup, we could've added secondary inits to `CoreRequestContext`. Not sure if that's a fair concern though

adam-fowler:
> I like this a lot. My only "concern" is that we lose the ability to add more properties to `CoreRequestContext` later with this setup. With the old setup, we could've added secondary inits to `CoreRequestContext`. Not sure if that's a fair concern though

Not sure how this affects your ability to write a secondary init for `CoreRequestContext`. You can still do that and call it from your custom context.

Joannis (approved):

PR #471, opened 2024-06-11
Make `chunkSize` for file reads configurable and default to the suggested value from `SwiftNIO`

### LINKED ISSUE #470: 2.x: Slow troughput for `FileIO.readFile()`
Using [`FileIO.readFile()`](https://github.com/hummingbird-project/hummingbird/blob/ea60c9b2d66ddb27e2aa06982a1c5477f9a50b83/Sources/Hummingbird/Files/FileIO.swift#L106) seems to be much slower than `FileIO.writeFile()` and CPU usage is very high. This has the effect that, on my machine, downloading large files is limited to ~50MB/s while uploading is possible at 1GB/s or more.

I've noticed the following:
- A larger chunk size increases the throughput considerably
- The relevant `chunkSize` is defined in `readFile()` as `8 * 1024` (so 8 kB)
- The `chunkSize` defined in the struct `FileIO` is not used at all (but would read from `NonBlockingFileIO.defaultChunkSize`, which is much larger at 128 kB)

Suggestions:
- Make `chunkSize` an optional parameter for `loadFile()` and `readFile()`, defaulting to `NonBlockingFileIO.defaultChunkSize`
- Remove `chunkSize` in `FileIO`

I can make a pull request with these changes if you find them reasonable.

PS: Thanks for this great library!

andreasley:
This PR lowers the CPU usage for file reads in `FileIO` by increasing the `chunkSize` to SwiftNIO's default value (`NonBlockingFileIO.defaultChunkSize`).

It also makes the `chunkSize` configurable for each read.

Fixes #470

### DIFF
--- a/Sources/Hummingbird/Files/FileIO.swift
+++ b/Sources/Hummingbird/Files/FileIO.swift
@@ -34,12 +34,12 @@ public struct FileIO: Sendable {
     /// - Parameters:
     ///   - path: System file path
     ///   - context: Context this request is being called in
-    ///   - chunkSize: Size of the chunks read from disk and loaded into memory (in bytes). Defaults to the value suggested by `swift-nio`.
+    ///   - chunkLength: Size of the chunks read from disk and loaded into memory (in bytes). Defaults to the value suggested by `swift-nio`.
     /// - Returns: Response body
-    public func loadFile(path: String, context: some BaseRequestContext, chunkSize: Int = NonBlockingFileIO.defaultChunkSize) async throws -> ResponseBody {
+    public func loadFile(path: String, context: some BaseRequestContext, chunkLength: Int = NonBlockingFileIO.defaultChunkSize) async throws -> ResponseBody {
         do {
             let stat = try await fileIO.lstat(path: path)
-            return self.readFile(path: path, range: 0...numericCast(stat.st_size - 1), context: context, chunkSize: chunkSize)
+            return self.readFile(path: path, range: 0...numericCast(stat.st_size - 1), context: context, chunkLength: chunkLength)
         } catch {
             throw HTTPError(.notFound)
         }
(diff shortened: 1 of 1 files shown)


adam-fowler (commented):
Looks good, I'm just wondering, now that it is public, if there is a better name than chunkSize

andreasley:
> Looks good, I'm just wondering, now that it is public, if there is a better name than chunkSize

I was asking myself the same question. Something like `memoryBufferLimit` perhaps, but I'm not really convinced.

I also wasn't sure if `chunkSize` should be defined in the parent struct instead. From the way `Hummingbird.FileIO` is used in hummingbird-examples, I considered the current PR more flexible.

And finally, I wondered if FileIO should switch to the (not quite finished) `NIOFileSystem` instead of using `NonBlockingFileIO`.

adam-fowler:
It will move to NIOFileSystem but only once the API is locked down which it currently isn't. It would be good to verify you can provide a chunk size to the streaming APIs for NIOFileSystem before we merge this as well.

andreasley:
> It would be good to verify you can provide a chunk size to the streaming APIs for NIOFileSystem before we merge this as well.

That's possible. There is [ReadFileHandle.readChunks(in:chunkLength:)](https://swiftpackageindex.com/apple/swift-nio/2.66.0/documentation/_niofilesystem/readfilehandle/readchunks(in:chunklength:)), which returns an `AsyncSequence`.

adam-fowler:
@andreasley I guess use `chunkLength` given that is used NIOFileSystem. You also need to run Nick Lockwood's swiftformat on your changes (v0.51.15 or later)

andreasley:
Rename done.

`swiftformat` didn't make any corrections to `FileIO.swift`, the only file I've changed, but it's the first time I'm using it, so maybe I've missed something.

adam-fowler:
Seems to be completely unrelated code that is not formatted correctly. I fix that separately

adam-fowler (approved):

PR #475, opened 2024-06-15
Playing with naming: rename InstantiableRequestContext to InitializableFromSource and set Source as primary type 

adam-fowler:
So `Context: InstantiableRequestContext where Context.Source == Type` becomes `Context: InitializedFrom<Type>`

Also 
- renamed `ServerRequestContextSource` to `ApplicationRequestContextSource`
- moved some code about


Joannis:
I'd like this to become a bit more of a discussion before we make a bunch of PRs to change names.

adam-fowler:
The PR was meant to be a discussion. We can do this in another forum if you want. I'm not 100% on the name. I think the primary associated type works quite well.

Sources/Hummingbird/Utils/InitializedFromSource.swift
```diff
@@ -0,0 +1,20 @@
...
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+/// A type that can be initialized from another type
+public protocol InitializableFromSource<Source>: Sendable {
```
Joannis, line 16: Filename needs to change to `InitializableFromSource.swift`

Joannis (approved):

PR #479, opened 2024-06-18
Router Bugfixes

Joannis:


Joannis:
This PR moves a lot of logic into a new struct. I don't know if it has a performance impact yet, so I might need to undo that

Sources/Hummingbird/Router/Trie/Trie+resolve.swift
```diff
@@ -18,130 +18,197 @@ extension RouterTrie {
...
-                node: node,
-                parameters: &parameters
-            )
-            switch result {
-            case .match, .deadEnd:
-                nodeIndex += 1
+                // Keep track of the nextNodeIndex known locally
+                // As we descend, this value changes. However, if descending doesn't yield a result
+                // We'll want to continue on the next path
+                let lastKnownNextNodeIndex = nextNodeIndex
```
adam-fowler, line 114: This is not used

Sources/Hummingbird/Router/Trie/Trie+resolve.swift
```diff
@@ -18,130 +18,197 @@ extension RouterTrie {
...
-        }
 
-        if let value = self.values[node.valueIndex] {
-            return (value: value, parameters: parameters)
-        } else {
-            return nil
-        }
-    }
+            if let nextComponent = nextPathComponent(advancingIndex: &nextPathComponentIndex) {
+                // There's another component to the route
```
adam-fowler, line 109: We can do this because the trie nodes are ordered in priority?

Sources/Hummingbird/Router/Trie/Trie+resolve.swift
```diff
@@ -18,130 +18,197 @@ extension RouterTrie {
...
+            } else {
+                return nil
+            }
+        }
+
+        @inlinable
+        mutating func descend(
+            component: Substring,
+            nextPathComponentIndex: Int,
+            nextNodeIndex: inout Int
```
adam-fowler, line 80: `nextNodeIndex` seems a little bit of a misnomer here. Given you are matching `component` shouldn't it just be `nodeIndex`.

I understand why `nextPathComponentIndex` had the next prefix as in it is the index of the next path component after `component`, but not sure it makes sense for `nextNodeIndex`.

adam-fowler (commented):
In general looks good, frustrating it requires the recursive function. 
You will need to do some benchmarks with this to ensure it doesn't cost too much

Sources/Hummingbird/Router/Trie/Trie+resolve.swift
Joannis: Depends a bit on where in the code's flow you're currently at whether or not it's the current component or the next one.

Sources/Hummingbird/Router/Trie/Trie+resolve.swift
Joannis: Correct.

adam-fowler (approved):

PR #482, opened 2024-06-20
File middleware should set endpoint if it hits a file

adam-fowler:


Joannis (approved):
Good find

adam-fowler:
I've now changed this, to only report FileMiddleware as the source. I was told FileMiddleware should probably not use the file name as a span name, it has a high cardinality problem and exposes too much information. 

I have also added setting the endpoint path as you enter a RouterGroup, so if a middleware throws an error you have some router info

Joannis (approved):

PR #484, opened 2024-06-24
Fix HTTP error serialization

Joannis:
Before this PR, HTTPError would not apply escaping to characters inside the message String

### DIFF
--- a/Sources/Hummingbird/Error/HTTPError.swift
+++ b/Sources/Hummingbird/Error/HTTPError.swift
@@ -13,7 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 import HTTPTypes
+import Foundation
 import NIOCore
+import NIOFoundationCompat
 
 /// Default HTTP error. Provides an HTTP status and a message
 public struct HTTPError: Error, HTTPResponseError, Sendable {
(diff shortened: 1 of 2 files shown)


adam-fowler:
Good catch, but I'm wondering if we should change this slightly. `HTTPError` assumes we are outputting json. Maybe we should use the encoder from the context to build the error. We can do this now `HTTPError` is defined in Hummingbird and not HummingbirdCore.

Sources/HummingbirdRouter/RouterBuilder.swift
```diff
@@ -46,8 +46,12 @@ public struct RouterBuilder<Context: RouterRequestContext, Handler: MiddlewarePr
 /// extend Router to conform to Responder so we can use it to process `Request``
 extension RouterBuilder: HTTPResponder, HTTPResponderBuilder {
     public func respond(to request: Input, context: Context) async throws -> Output {
-        try await self.handle(request, context: context) { _, _ in
-            throw HTTPError(.notFound)
+        do {
+            return try await self.handle(request, context: context) { _, _ in
+                throw HTTPError(.notFound)
```
adam-fowler, line 51: Given this will be caught straight is it worthwhile just returning a `Response(status: .notFound)`

adam-fowler (changes requested):
I can't see anyway we can get away without the status and headers in the HTTPErrorResponse. Looks good. There are a couple of minor things first before we merge.

PR #486, opened 2024-06-24
Remove deprecated symbols

adam-fowler:


Joannis (approved):

MahdiBM (commented):
Another option is not to remove these, but instead mark them as unavailable.
That'll enable you to remove them whenever you want, but it'll also make sure the symbols are there for diagnosis for whoever finally updates their hummingbird.
It doesn't really cost Hummingbird but will make sure users have an even better experience updating even if they're kind of late on updating their packages.

adam-fowler:
> Another option is not to remove these, but instead mark them as unavailable. That'll enable you to remove them whenever you want, but it'll also make sure the symbols are there for diagnosis for whoever finally updates their hummingbird. It doesn't really cost Hummingbird but will make sure users have an even better experience updating even if they're kind of late on updating their packages.

ok makes sense

PR #487, opened 2024-06-25
Add FileIdentifier associated type to FileProvider

adam-fowler:
Allow for FileProviders that may not use Strings to identify their files

### DIFF
--- a/Sources/Hummingbird/Files/FileProvider.swift
+++ b/Sources/Hummingbird/Files/FileProvider.swift
@@ -25,14 +25,7 @@ public protocol FileProvider: Sendable {
     /// Get file identifier
     /// - Parameter path: path from URI
     /// - Returns: File Identifier
-    func getFileIdentifier(_ path: String) throws -> FileIdentifier
-
-    /// Append a file name component to a file identifier
-    /// - Parameter
-    ///   - filename: File name to append
-    ///   - path: File Identifier
-    /// - Returns: Resulting file identifier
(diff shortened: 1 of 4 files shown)


Sources/Hummingbird/Files/FileProvider.swift
```diff
@@ -19,29 +19,31 @@ import NIOPosix
 public protocol FileProvider: Sendable {
     /// File attributes type
     associatedtype FileAttributes
+    /// File identifier
+    associatedtype FileIdentifier
```
MahdiBM, line 23: I understand that this is not needed currently, but i think for the sake of future API flexibility, it's reasonable to require this associated type to be `Hashable`.
For example maybe someday in the future you want to add a new function to the protocol but provide a default implementation to prevent API breakages. `Hashable` could help in a day like that.

adam-fowler: I cannot see any situation where this would need to be `Hashable` in the protocol. The one situation where this is used currently is the FileMiddleware and that only needs to pass it around to identify a file. If it was used elsewhere in the future we could always add the requirement `where FileProvider.FileIdentifier: Hashable`.

I can see situations where specific implementations of FileProviders would require it to `Hashable` eg A FileProvider that caches files in memory and needs a lookup, but that requirement would be implemented in the implementation and not the protocol.

MahdiBM: I'm convinced, you're right 🙂.

Joannis (approved):

PR #491, opened 2024-06-27
Use RouterPath instead of String as parameter in `RouterMethods.on`

adam-fowler:


Sources/Hummingbird/Router/RouterPath.swift
```diff
@@ -14,19 +14,88 @@
...
-        case capture(Substring)
-        case prefixCapture(suffix: Substring, parameter: Substring) // *.jpg
-        case suffixCapture(prefix: Substring, parameter: Substring) // file.*
-        case wildcard
-        case prefixWildcard(Substring) // *.jpg
-        case suffixWildcard(Substring) // file.*
-        case recursiveWildcard
-        case null
+    public struct Element: Equatable, Sendable, CustomStringConvertible {
+        package enum _Internal: Equatable, Sendable {
```
Joannis, line 18: I think this needs to be `@_spi(Internal) public` for the benchmarks

Joannis (approved):
LGTM, but let's make sure this doesn't break the benchmarks - or that we've fixed the benchmarks. Approving already since I don't want to hold up the merge for a sporadically used benchmark

adam-fowler:
> LGTM, but let's make sure this doesn't break the benchmarks - or that we've fixed the benchmarks. Approving already since I don't want to hold up the merge for a sporadically used benchmark

I've checked against wrk and there is no change in performance. I can't get the benchmarks to produce any consistent data at the moment so didn't verify against them

PR #496, opened 2024-06-29
Enable the builder pattern on MiddlewareGroup

runhum:
- Makes the `MiddlewareGroup.add(_ middleware:)` function return itself

This enables the following API:

```swift
let router = Router()

router.middlewares
    .add(LogRequestsMiddleware(.info))
    .add(MyMiddleware())
    .add(...)

```


adam-fowler (approved):
Looks good

PR #498, opened 2024-07-01
Make MediaType.Category values public again

CallumTodd7:
#485 (accidentally?) set the MediaType.Category values to internal visibility. This PR reverts that change to avoid the need for `.init(rawValue: "image")!`, etc.


adam-fowler (approved):
🤦 
Cheers

Issue #501, opened 2024-07-03
HTTPResponseError is now a ResponseGenerator


sidepelican:
After this PR, https://github.com/hummingbird-project/hummingbird/pull/484

This PR has made `HTTPResponseError` a confusing entity.
Although `status` and `headers` are included as requirements, but there are no meaningful usages.
It is no different from `ResponseGenerator`.

To avoid confusion, we should either remove it, or provide a default conformance to `ResponseGenerator`.

adam-fowler:
`status` and `headers` are used in both the tracing and metrics middleware, so we can't just conform to ResponseGenerator as we need this additional information.

sidepelican:
Sorry, I missed the usage in TracingMiddleware. My thoughts are written here: https://github.com/hummingbird-project/hummingbird/pull/502#issuecomment-2207866751

PR #502, opened 2024-07-03
Add default conformance of ResponseGenerator for HTTPResponseError

sidepelican:
This is one of the solution of #501.

Add default conformance of ResponseGenerator for HTTPResponseError


adam-fowler (commented):
This could be an alternative to what we have. I’ll need to have a think about it.

sidepelican:
My hope is that `HTTPResponseError` will become a utility to easily convert its own errors into a Response, similar to `Vapor.AbortError`.

If ideologically it is not right to do so, I can accept that.
However, `HTTPResponseError` is confusing, so I would prefer it not to exist.

If you stop converting `HTTPResponseError` in `RouterResponder` and convert it in a new Middleware instead, there would be no need to handle `HTTPResponseError` in `TracingMiddleware` (Sorry this is just idea).

adam-fowler:
> My hope is that `HTTPResponseError` will become a utility to easily convert its own errors into a Response, similar to `Vapor.AbortError`.
> 
> If ideologically it is not right to do so, I can accept that. However, `HTTPResponseError` is confusing, so I would prefer it not to exist.
> 
> If you stop converting `HTTPResponseError` in `RouterResponder` and convert it in a new Middleware instead, there would be no need to handle `HTTPResponseError` in `TracingMiddleware` (Sorry this is just idea).

HTTPResponseError is a protocol you conform your error to, if you want to generate a response from that error. It doesn't require a definition for the error body as it could come in many forms, a string, buffer, object. How you generate a response from that is defined in your implementation of `response(from:context)`. Also the response generated by the error should be in the format expected. If the response encoder outputs XML then the error should be in XML, so you need the request headers (at least the accept header) and request context to be able to generate the correct payload and augment the headers with content-type.

sidepelican:
Ok. HTTPResponseError is not the utility I thought it was. I can agree with the design approach, so I will close this PR. Thank you.

PR #510, opened 2024-07-11
Add option to remove prefix from path in FileMiddleware

adam-fowler:


sliemeobn:
looks good. maybe the name is a bit awkward?

do you think just using `basePath` is too terse? I feel it is a common term for this kind of thing.

adam-fowler:
> do you think just using `basePath` is too terse? I feel it is a common term for this kind of thing.

What does `basePath` mean when you also have `rootFolder`. In actual fact it should probably be `baseFolderToRemove` to make it clear it will only remove folders and not incomplete path elements.

sliemeobn:
> @sliemeobn what about `urlBaseFolder`

the second I read 'folder' I think about file system, but maybe that's just me.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -95,7 +102,7 @@ public struct FileMiddleware<Context: RequestContext, Provider: FileProvider>: R
             }
 
             // Remove percent encoding from URI path
-            guard let path = request.uri.path.removingPercentEncoding else {
+            guard var path = request.uri.path.removingPercentEncoding else {
```
sliemeobn, line 105: I think before we do anything, we should actually check if the path start with the pathPrefix.
this way can mount several FileMiddlewares on different URL paths (eg: each service a separate file system directoy)

sliemeobn: also, a quick-bail if the method is not `get` or `head` would be a good idea I think

adam-fowler: I added the quick bail for get and head. Also added bailing on path prefix not existing. But this needs to be done after the percent encoding is removed. Test is updated to include 2 middleware which return slightly different results

Joannis (changes requested):
I agree that `urlBasePath` needs a different name. Looks good otherwise

adam-fowler:
> I agree that `urlBasePath` needs a different name. Looks good otherwise

So far we've had `pathPrefixToRemove`, `basePath` and `urlBasePath`. I would prefer we keep the `url` in the name to indicate which path we talking about. What about a simple `urlPrefix`.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -94,8 +101,12 @@ public struct FileMiddleware<Context: RequestContext, Provider: FileProvider>: R
                 throw error
             }
 
+            guard request.method == .get || request.method == .head else {
+                throw error
```
Joannis, line 105: Do we need to throw an error, or can we forward to the next handler?

adam-fowler:
> What about `routePrefix`?

Think I prefer `url` over `route`

Sources/Hummingbird/Middleware/FileMiddleware.swift
adam-fowler: We've already run the next handler. The file middleware only looks for files if the response is a 404

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -44,24 +44,28 @@ public protocol FileMiddlewareFileAttributes {
...
     /// - Parameters:
     ///   - rootFolder: Root folder to look for files
+    ///   - urlBasePath: Prefix to remove from request URL
     ///   - cacheControl: What cache control headers to include in response
     ///   - searchForIndexHtml: Should we look for index.html in folders
     ///   - threadPool: ThreadPool used by file loading
     ///   - logger: Logger used to output file information
     public init(
         _ rootFolder: String = "public",
+        urlBasePath: String? = nil,
```
Joannis, line 60: One more question: Can we change `urlBasePath` to use `RouterPath` instead? That would clarify the API a lot more for me.

adam-fowler: > One more question: Can we change `urlBasePath` to use `RouterPath` instead? That would clarify the API a lot more for me.

Making it a RouterPath makes sense

adam-fowler: Actually it doesn't make sense. Using a router path implies we support wildcards, parameters when loading files

Joannis:
Idea from a DM:
- Maybe we could make the middleware aware of what route path it's been added on. That way we wouldn't pass it in here, but at most add a boolean to remove that path's prefix

adam-fowler:
> Idea from a DM:
> 
> - Maybe we could make the middleware aware of what route path it's been added on. That way we wouldn't pass it in here, but at most add a boolean to remove that path's prefix

This won't work if the middleware is a reference type and is added in different places in the router.

PR #513, opened 2024-07-15
Transforming router group

adam-fowler:
Create a TransformingRouterGroup that transforms the request context. If a route is matched in this group then the request context is transformed as it enters the group middleware chain.

For a RequestContext to be convertible from the original RequestContext it needs to set its `InitializableFromSource.Source` to the original context. To get this to work I had to extend `RequestContext` to conform to `RequestContextSource`.

I had to change how groups worked internally to get this to work. Instead of collating all the middleware that should be applied to endpoints in a group and creating a middleware responder and adding that directly to the router , it keeps a reference to the parent of the group and adds the middleware responder to the parent.

I remove `group` function requirement from `RouterMethods` and replaced it with an implementation in the extension of `RouterMethods` as all version of this function are the same.

Also implemented this for the RouterBuilder with `TransformingRouteGroup`.


Sources/Hummingbird/Router/TransformingRouterGroup.swift
```diff
@@ -0,0 +1,87 @@
...
+/// The new Context's ``RequestContext.Source`` needs to be he context of the group parent so the
+/// the new context can be initialized from the context of the parent.
+/// ```
+/// router
+/// .transformingGroup("todos", context: MyContext.self)
+/// .get(use: todoController.list)
+/// .put(use: todoController.create)
+/// .get(":id", use: todoController.get)
+/// .put(":id", use: todoController.update)
+/// .delete(":id", use: todoController.delete)
```
Joannis, line 34:
```suggestion
///   .transformingGroup("todos", context: MyContext.self)
///   .get(use: todoController.list)
///   .put(use: todoController.create)
///   .get(":id", use: todoController.get)
///   .put(":id", use: todoController.update)
///   .delete(":id", use: todoController.delete)
```

Joannis (changes requested):
Let's discuss a new name for `func transformingGroup`. Other than that, this looks perfect

adam-fowler:
> Let's discuss a new name for `func transformingGroup`. Other than that, this looks perfect

Ok we now have `RouterMethods.group(_:context:)`. Also I've hidden the TransformingRouterGroup type

Sources/Hummingbird/Router/TransformingRouterGroup.swift
```diff
@@ -0,0 +1,59 @@
...
+        }
+    }
+
+    init(parent: any RouterMethods<InputContext>) {
+        self.parent = parent
+    }
+
+    /// Add middleware (Stub function as it isn't used)
+    @discardableResult func add(middleware: any RouterMiddleware<Context>) -> Self {
+        preconditionFailure("Cannot add middleware to context transform")
```
adam-fowler, line 40: I could add a separate protocol for Router types that can be a parent but don't hold middleware. To avoid adding the runtime error.

Joannis (approved):

adam-fowler:
For the RouterBuilder I have separated out the context transform into `ContextTransform` so people would write it as follows. What are your thoughts?
```swift
let router = RouterBuilder(context: BasicRouterRequestContext.self) {
    RouteGroup("/group") {
        ContextTransform(context: TestRouterContext2.self) {
            TestTransformMiddleware()
            Get { _, context in
                return Response(status: .ok, headers: [.middleware2: context.string])
            }
        }
    }
}
```

Sources/Hummingbird/Router/TransformingRouterGroup.swift
adam-fowler: I decided this isn't worth it. `TransformingRouterGroup` is a type the user will never get their hands on, so we can leave with the runtime error

Joannis:
I love it.

Joannis (approved):

PR #518, opened 2024-08-06
Max Connection manager

adam-fowler:
Channel handler limiting the number of active connections

Currently just based on number of connections, but we could add a delegate to make this more flexible


Sources/Hummingbird/Configuration.swift
```diff
@@ -33,6 +33,8 @@ public struct ApplicationConfiguration: Sendable {
     public var backlog: Int
     /// Allows socket to be bound to an address that is already in use.
     public var reuseAddress: Bool
+    /// Maximum active connections
+    public let maxActiveConnections: Int?
```
Joannis, line 37: Since there are multiple ways to limit connections besides a hard cap, I'm a little reluctant to add such a property here

adam-fowler: This is still a draft and more a demonstration of how we can limit available connections. I was thinking of adding a `Delegate` generic parameter to the ChannelHandler which gave us a little more flexibility. eg
```swift
protocol AvailableConnectionsDelegate {
    func addConnection()
    func removeConnection()
    func IsConnectionAvailable() -> Bool
}
```

Joannis: That's roughly what I'm thinking too

Joannis:
Looks like a solid PR. Would we ever need to consider making this protocol support `async` returns?

Joannis:
It doesn't make sense to do a DB call here/to support that. But maybe actor isolated state can be relevant.

adam-fowler:
> Looks like a solid PR. Would we ever need to consider making this protocol support `async` returns?

It is being called from a ChannelHandler so supporting async is not a great idea. If you have go to another task before making a decision on whether to accept a connection that's going to kill performance.

Joannis (approved):
Looks good!

Sources/Hummingbird/Configuration.swift
```diff
@@ -109,6 +115,7 @@ public struct ApplicationConfiguration: Sendable {
             serverName: self.serverName,
             backlog: self.backlog,
             reuseAddress: self.reuseAddress,
+            availableConnectionDelegate: self.availableConnectionDelegate,
```
MahdiBM, line 118: inconsistent naming: `AvailableConnections` vs `availableConnection`

MahdiBM (commented):
Is it going to be worth the protocol if it doesn't even have any EL or async/await APIs?

The default implementation already accepts "maxConnextions" as a customization.

In the future a protocol can always be added if needed.

If you actually foresee situations where this will be helpful, then I have no objections.

adam-fowler:
> Is it going to be worth the protocol if it doesn't even have any EL or async/await APIs?
> 
> The default implementation already accepts "maxConnextions" as a customization and in the future a protocol can always be re-added if needed.
> 
> If you actually foresee situations where this will be helpful, then I have no objections.

There are other resources you might base it off eg memory

PR #519, opened 2024-08-09
Use a response writer at the Server level

adam-fowler:
This is an experiment

Instead of returning a Response to HTTPChannelHandler pass a ResponseWriter to the request handler. This change would not be propagated up the chain though, users would still be returning a Response from the Router. But Application would use the ResponseWriter to write the Response returned by the Router instead of passing it back to the HTTPChannelHandler.

Why would we do this? It gives us flexibility in the future to support things like 1xx headers, better support tracing maybe at the server level instead of the slightly flakey way we do it at the moment where we end the trace span in a non-structured way.

For most users this will have no visible effect.

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
@@ -12,6 +12,7 @@
 //
 //===----------------------------------------------------------------------===//
 
+import HTTPTypes
 import NIOCore
 
 /// HTTP Response Body part writer
@@ -22,6 +23,9 @@ public protocol ResponseBodyWriter {
     /// Write a sequence of ByteBuffers
     /// - Parameter buffers: Sequence of buffers
     func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
+    /// Finish writing body
+    /// - Parameter trailingHeaders: Any trailing headers you want to include at end
+    consuming func finish(_ trailingHeaders: HTTPFields?) async throws
 }
 
 extension ResponseBodyWriter {
@@ -30,6 +34,15 @@ extension ResponseBodyWriter {
     public func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws {
         for part in buffers {
             try await self.write(part)
+        }
+    }
+
+    ///  Write AsyncSequence of ByteBuffers
+    /// - Parameter buffers: ByteBuffer AsyncSequence
+    @inlinable
+    public func write<BufferSequence: AsyncSequence>(_ buffers: BufferSequence) async throws where BufferSequence.Element == ByteBuffer {
+        for try await buffer in buffers {
+            try await self.write(buffer)
         }
     }
 }
@@ -50,6 +63,12 @@ struct MappedResponseBodyWriter<ParentWriter: ResponseBodyWriter>: ResponseBodyW
         for part in buffers {
             try await self.parentWriter.write(self.transform(part))
         }
+    }
+
+    /// Finish writing body
+    /// - Parameter trailingHeaders: Any trailing headers you want to include at end
+    consuming func finish(_ trailingHeaders: HTTPFields?) async throws {
+        try await self.parentWriter.finish(trailingHeaders)
     }
 }
 
--- a/Sources/HummingbirdCore/Response/ResponseWriter.swift
+++ b/Sources/HummingbirdCore/Response/ResponseWriter.swift
@@ -17,7 +17,7 @@ import NIOCore
 import NIOHTTPTypes
 
 /// ResponseWriter that writes directly to AsyncChannel
-public struct ResponseWriter {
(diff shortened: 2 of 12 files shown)


Sources/HummingbirdCore/Response/ResponseWriter.swift
```diff
@@ -0,0 +1,47 @@
...
+    public func write(_ parts: some Sequence<HTTPResponsePart>) async throws {
+        for part in parts {
+            try await self.outbound.write(part)
+        }
+    }
+
+    ///  Write AsyncSequence of response parts
+    /// - Parameter parts: response parts AsyncSequence
+    @inlinable
+    public func write<AsyncSeq: AsyncSequence>(_ parts: AsyncSeq) async throws where AsyncSeq.Element == HTTPResponsePart {
```
Joannis, line 42: Nit:

```suggestion
    public func write<Parts: AsyncSequence>(_ parts: Parts) async throws where Parts.Element == HTTPResponsePart {
```

Joannis (commented):
I like this a bit more than what we did thus far.

Joannis:
We can consider adding some validation to writing the parts the ResponseWriter so the response format is correct. But I do think NIO already have that.

adam-fowler:
> We can consider adding some validation to writing the parts the ResponseWriter so the response format is correct. But I do think NIO already have that.

NIO have that at the runtime level, but being able to force it at compile time would be good. I haven't thought of an easy way to implement that though.

Sources/HummingbirdCore/Response/ResponseWriter.swift
```diff
@@ -0,0 +1,87 @@
...
+        return RootResponseBodyWriter(outbound: self.outbound)
+    }
+
+    /// Write Informational HTTP head part
+    ///
+    /// Calling this with a non informational HTTP response head will cause a precondition error
+    /// - Parameter head: Informational response head
+    @inlinable
+    public func writeInformationalHead(_ head: HTTPResponse) async throws {
+        precondition((100..<200).contains(head.status.code), "Informational HTTP responses require a status code between 100 and 199")
```
Joannis, line 40:
```suggestion
        precondition((100..<200).contains(head.status.code), "Informational HTTP responses require a status code in the range of 100 through 199")
```

Sources/HummingbirdCore/Response/ResponseWriter.swift
```diff
@@ -0,0 +1,87 @@
...
+        try await self.outbound.write(.head(head))
+        return RootResponseBodyWriter(outbound: self.outbound)
+    }
+
+    /// Write Informational HTTP head part
+    ///
+    /// Calling this with a non informational HTTP response head will cause a precondition error
+    /// - Parameter head: Informational response head
+    @inlinable
+    public func writeInformationalHead(_ head: HTTPResponse) async throws {
```
Joannis, line 39: Is this not consuming?

Tests/HummingbirdCoreTests/CoreTests.swift
```diff
@@ -159,29 +161,19 @@ class HummingBirdCoreTests: XCTestCase {
...
-            XCTAssertEqual(response.status, .unauthorized)
-            XCTAssertEqual(response.headers[.contentLength], "0")
-        }
-    }
-
     func testConsumeBody() async throws {
         try await testServer(
-            responder: { request, _ in
+            responder: { (request, responseWriter: consuming ResponseWriter, _) in
+                let buffer: ByteBuffer
```
Joannis, line 167: What a great catch thanks to Non-Copyables

Sources/HummingbirdCore/Response/ResponseWriter.swift
adam-fowler: No, you need to be able to send the actual header after the informational header.

Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
```diff
@@ -22,6 +23,9 @@ public protocol ResponseBodyWriter {
     /// Write a sequence of ByteBuffers
     /// - Parameter buffers: Sequence of buffers
     func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
+    /// Finish writing body
+    /// - Parameter trailingHeaders: Any trailing headers you want to include at end
+    consuming func finish(_ trailingHeaders: HTTPFields?) async throws
```
Joannis, line 28: AFAIK, a `consuming func` is meaningless on a copyable type

Joannis (commented):
LGTM, but I wonder if we should make ResponseBodyWriter `~Copyable`

adam-fowler:
> LGTM, but I wonder if we should make ResponseBodyWriter `~Copyable`

Would love to but we can't add `~Copyable` on a protocol.

Joannis (approved):
Let's fix the formatting issues for the CI and we're good to go

PR #529, opened 2024-08-26
Make ResponseBodyWriter methods mutating

adam-fowler:
This means a `ResponseBodyWriter` with internal mutable state isn't forced to be reference object.

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBody.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBody.swift
@@ -18,7 +18,7 @@ import NIOCore
 /// Response body
 public struct ResponseBody: Sendable {
     @usableFromInline
-    let _write: @Sendable (any ResponseBodyWriter) async throws -> Void
+    let _write: @Sendable (inout any ResponseBodyWriter) async throws -> Void
     public let contentLength: Int?
 
     /// Initialise ResponseBody with closure writing body contents.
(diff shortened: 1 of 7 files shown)


Sources/HummingbirdCore/Response/ResponseBody.swift
```diff
@@ -70,8 +70,8 @@ public struct ResponseBody: Sendable {
     }
 
     @inlinable
-    public consuming func write(_ writer: any ResponseBodyWriter) async throws {
-        try await self._write(writer)
+    public consuming func write(_ writer: consuming any ResponseBodyWriter) async throws {
```
Joannis, line 73: Isn't this supposed to be `inout any ResponseBodyWriter`?

Joannis (approved):

Joannis:
I think the PR has a typo, but then the code and tests run fine so I'm not sure that the typo I found is a typo at all. I'll approve it since it 'works', but please do check that first

Sources/HummingbirdCore/Response/ResponseBody.swift
adam-fowler: In the situations where I could use this as an inout ie tests it doesn't work as I have to pass it as `any ResponseBodyWriter`. You can't pass a concrete version of the protocol as the inout allows you to change the type to any `ResponseBodyWriter`. 

Because the variable is consumied it is mutable within the function that consumed it.

PR #531, opened 2024-08-26
Fix serving files

adam-fowler:


Sources/Hummingbird/Files/FileIO.swift
```diff
@@ -123,6 +123,7 @@ public struct FileIO: Sendable {
                     fileOffset = range.index(fileOffset, offsetBy: bytesToRead)
                     try await writer.write(buffer)
                 }
+                try await writer.finish(nil)
```
Joannis, line 126: Is this something we can test in a unit test?

Joannis (approved):

Sources/Hummingbird/Files/FileIO.swift
adam-fowler: I thought we did, obviously not. I'll add a test in a separate PR.

PR #532, opened 2024-08-26
Return empty `ResponseBody` in `FileIO` for empty files

andreasley:
The FileIO methods for loading files create byte ranges based on the file's size. If the file is empty, the range cannot be created and the app exits with the following error:

```error
Fatal error: Range requires lowerBound <= upperBound
```

This pull request adds a check to ensure the filesize is > 0 and just returns an empty `ResponseBody` otherwise.


adam-fowler:
Thank you. I'll look to get a release out tomorrow with these fixes.

PR #533, opened 2024-08-27
Remove async from Environment.dotEnv()

JacobHearst:
`Environment.dotEnv(_:)` was marked `async` despite not performing any async actions. This was complicating my attempts to extract `Environment` as a dependency with [swift-dependencies](https://github.com/pointfreeco/swift-dependencies) as that library assumes that dependencies can be synchronously initialized.

As far as I can tell, there's no need for `dotEnv(_:)` to be `async` so I've removed the superfluous keywords.

### DIFF
--- a/Sources/Hummingbird/Environment.swift
+++ b/Sources/Hummingbird/Environment.swift
@@ -127,13 +127,13 @@ public struct Environment: Sendable, Decodable, ExpressibleByDictionaryLiteral {
     }
 
     /// Create Environment initialised from the `.env` file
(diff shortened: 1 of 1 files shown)


adam-fowler:
loading the dotEnv file is a blocking operation. We will be moving its loading to NIOFileSystem once NIO finalise the API and at that point it will be an async function.

JacobHearst:
> loading the dotEnv file is a blocking operation. We will be moving its loading to NIOFileSystem once NIO finalise the API and at that point it will be an async function.

Good to know, thanks for the quick response!

PR #537, opened 2024-09-05
Type erasing ResponseBodyWriter

adam-fowler:
- Rename ResponseBodyWriter to ResponseBodyWriterProtocol
- Add type ResponseBodyWriter that holds an existential ResponseBodyWriterProtocol
- Broke code into three files to make it easier to read

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
@@ -15,67 +15,24 @@
 import HTTPTypes
 import NIOCore
 
-/// HTTP Response Body part writer
-public protocol ResponseBodyWriter {
-    /// Write a single ByteBuffer
-    /// - Parameter buffer: single buffer to write
-    mutating func write(_ buffer: ByteBuffer) async throws
-    /// Write a sequence of ByteBuffers
-    /// - Parameter buffers: Sequence of buffers
-    mutating func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
-    /// Finish writing body
-    /// - Parameter trailingHeaders: Any trailing headers you want to include at end
-    consuming func finish(_ trailingHeaders: HTTPFields?) async throws
-}
+/// Type erasing ``ResponseBodyWriterProtocol``.
+///
+/// Holds a pointer to any ResponseBodyWriterProtocol.
+public struct ResponseBodyWriter: ResponseBodyWriterProtocol {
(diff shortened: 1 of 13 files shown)


sliemeobn (commented):
nice!

I think this provides a nicer API for "users" of `ReponseBody` and more ergonomic extensions of it (ie: `ResponseBodyWriter` can now conform to protocols)

not 100% sure I like the naming of `ResponseBodyWriterProtocol`. this will only affect implementers of a ResponseBodyWriter, but it might be potentially confusing.

in the olden days words like `ResponseBodyWriting` were used, but it seems this is falling out of favor 🤔

Joannis (changes requested):
I understand the desire for this change, and I like it a lot. But I don't think that the value provided by this change outweighs a breaking change so late in the RC phase.

I'm happy to be convinced otherwise, maybe if more people weigh in

adam-fowler:
Another option is I ditch all the API breaking changes and simply add a `AnyResponseBodyWriter` type which the user can use in their own code.

Sources/HummingbirdCore/Response/ResponseBodyWriter.swift
```diff
@@ -15,67 +15,24 @@
...
-    mutating func write(contentsOf buffers: some Sequence<ByteBuffer>) async throws
-    /// Finish writing body
-    /// - Parameter trailingHeaders: Any trailing headers you want to include at end
-    consuming func finish(_ trailingHeaders: HTTPFields?) async throws
-}
+/// Type erasing ``ResponseBodyWriterProtocol``.
+///
+/// Holds a pointer to any ResponseBodyWriterProtocol.
+public struct ResponseBodyWriter: ResponseBodyWriterProtocol {
+    public var wrapped: any ResponseBodyWriterProtocol
```
MahdiBM, line 22: I wonder if sometime in the future you'd want to make an special case for hummingbird's own response-body-writer with an enum, for performance optimizations.

Just something to keep in mind. This public var shouldn't stop use from doing that.

MahdiBM: (Like `NIOAny`, or even String's own "guts")

adam-fowler:
Both protocol and type erased type need to be available. Users need to be able to write their own Response body writers. This new type fulfills the need for a concrete type that can be extended.

adam-fowler:
I've decided to not go ahead with this PR.

Issue #17, opened 2021-01-27
URLEncodedForm: Parse array indices


adam-fowler:
Currently arrays are formatted as follows
```
value[]=1&value[]=2
```
Add possibility of parsing arrays in form
```
value[0]=1&value[1]=2
```

supersonicbyte:
Hey! Is this still something that needs to be implemented?

adam-fowler:
> Hey! Is this still something that needs to be implemented?

I wrote this issue three years ago, I can't remember offhand what the context was. There certainly hasn't been any work on it since then.

The standard for arrays in url encoded form data, doesn't seem very clear but both of the above are valid. So we should probably support both. If you want to implement the parsing for the second format that'd be great

supersonicbyte:
Sure, I would love to try to implement the parsing.

PR #542, opened 2024-09-08
Add support for array indices in query params decoding

supersonicbyte:
This PR tries to resolve the following [issue](https://github.com/hummingbird-project/hummingbird/issues/17).

With the proposed implementation hummingbird will be able to decode arrays which contain indices from query parameters.

e.g.
`arr[0]=23&arr[1]=34&arr[2]=4`

The proposed implementation only allows such arrays if the indices are sent in an array of integers which increase by 1, meaning the indices must come in the strict order of `0, 1, 2, 3..` etc. Also we allow only ASCII numbers to be used as indices hence the need for custom extension for `Character`.
Otherwise the decoding will fail.


Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedFormNode.swift
```diff
@@ -84,6 +84,11 @@ enum URLEncodedFormNode: CustomStringConvertible, Equatable {
                 // currently don't support arrays and maps inside arrays
                 throw Error.notSupported
             }
+        case (.array(let array), .arrayWithIndices(let index)):
+            guard keys.count == 0, array.values.count == index else {
+                throw Error.notSupported
```
adam-fowler, line 89: Maybe we should add a new error here. `Error.invalidArrayIndex(Int)`

Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedFormNode.swift
```diff
@@ -186,8 +191,25 @@ enum KeyParser {
...
             if key[index] == "]" {
-                values.append(.array)
+                if let arrayIndex {
+                    values.append(.arrayWithIndices(arrayIndex))
+                } else {
+                    values.append(.array)
+                }
                 index = key.index(after: index)
             } else {
                 // an open bracket is unexpected
```
adam-fowler, line 215: Instead of parsing the array index above you should do it here once we have realised the next character isn't a `]`. You could also replace the `values.append(.map(key[index..<bracketIndex]))` call below with the following to do the integer conversion.
```swift
// If key can convert to an integer assume it is an array index
if let index = Int(key[index..<bracketIndex]) {
    values.append(.arrayWithIndices(index))
} else {
    values.append(.map(key[index..<bracketIndex]))
}
```

Tests/HummingbirdTests/URLEncodedForm/URLDecoderTests.swift
```diff
@@ -79,6 +79,41 @@ class URLDecodedFormDecoderTests: XCTestCase {
...
+    }
+
+    func testArrayWithIndicesThrows() {
+        struct Test: Codable, Equatable {
+            let arr: [Int]
+        }
+        let decoder = URLEncodedFormDecoder()
+        // incorrect indices
+        let query = "arr[0]=2?arr[2]=4"
+        XCTAssertThrowsError(try decoder.decode(Test.self, from: query))
```
adam-fowler, line 114: You should verify what error is thrown here.

adam-fowler (changes requested):
In general looks good. I've added some comments about decoding the integers.

supersonicbyte:
Thanks @adam-fowler for the review. I updated the PR, so please take a look when you find the time.

adam-fowler (approved):
This looks great, thank you

PR #552, opened 2024-09-14
Add verify documentation action

adam-fowler:
This is only to check it works. It will move to the project template after we have verified it works


Joannis:
Can this be closed now?

adam-fowler:
done

PR #561, opened 2024-09-26
Additional init for HTTPError, fix content type

adam-fowler:
Add new init
Removed the automatic addition of JSON content type, this is set by the response encoder


Joannis (approved):
Great catch!

PR #569, opened 2024-09-29
Check for HTTPResponseError in FileMiddleware, rather than a concrete type

Joannis:


Tests/HummingbirdTests/FileMiddlewareTests.swift
```diff
@@ -316,6 +316,55 @@ final class FileMiddlewareTests: XCTestCase {
         }
     }
 
+    func testOnThrow404() async throws {
```
adam-fowler, line 319: Not sure this test is required as all the other FileMiddleware tests are relying on this. 

The second test of course is necessary.

adam-fowler (changes requested):

adam-fowler:
Otherwise looks good. Just need to run swiftformat

adam-fowler (approved):

PR #572, opened 2024-09-30
Remove async from dotEnv.

connor-ricks:
The `dotEnv` function doesn't contain any asynchronous operations. Removing it allows consumers to load their environment in a non-asynchronous context.


adam-fowler:
At this point it doesn't but it contains a blocking operation (file loading) which will be replaced with NIOFilesystem calls when SwiftNIO finally do a proper release of it and will need to be async then.

PR #573, opened 2024-09-30
Use Foundation.ProcessInfo for environment parsing

### LINKED ISSUE #571: Use ProcessInfo.processInfo.environment instead of environ.
We are using Foundation in Hummingbird now so this shouldn't be an issue. It'll also allow us to get around the static sdk issues with `environ`.

GNMoseke:
Resolves #571 

- Simply replaces use of `environ` with `Foundation.ProcessInfo`

### DIFF
--- a/Sources/Hummingbird/Environment.swift
+++ b/Sources/Hummingbird/Environment.swift
@@ -112,8 +112,8 @@ public struct Environment: Sendable, Decodable, ExpressibleByDictionaryLiteral {
     /// Construct environment variable map
     static func getEnvironment() -> [String: String] {
         var values: [String: String] = [:]
-        for key in ProcessInfo.processInfo.environment.keys {
-            values[key.lowercased()] = ProcessInfo.processInfo.environment[key]
+        for item in ProcessInfo.processInfo.environment {
+            values[item.key.lowercased()] = item.value
         }
         return values
     }
--- a/Sources/Hummingbird/Environment.swift
+++ b/Sources/Hummingbird/Environment.swift
@@ -21,6 +21,7 @@ import Darwin.C
 #else
 #error("Unsupported platform")
 #endif
+import Foundation
 import HummingbirdCore
 import NIOCore
 
@@ -111,22 +112,8 @@ public struct Environment: Sendable, Decodable, ExpressibleByDictionaryLiteral {
     /// Construct environment variable map
     static func getEnvironment() -> [String: String] {
         var values: [String: String] = [:]
-        let equalSign = Character("=")
-        #if canImport(Musl)
-        guard let envp = environ else { return [:] }
-        #else
-        let envp = environ
-        #endif
-        var idx = 0
-
-        while let entry = envp.advanced(by: idx).pointee {
-            let entry = String(cString: entry)
-            if let i = entry.firstIndex(of: equalSign) {
-                let key = String(entry.prefix(upTo: i))
-                let value = String(entry.suffix(from: i).dropFirst())
-                values[key.lowercased()] = value
-            }
-            idx += 1
+        for key in ProcessInfo.processInfo.environment.keys {
+            values[key.lowercased()] = ProcessInfo.processInfo.environment[key]
         }
         return values
     }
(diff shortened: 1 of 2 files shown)


Joannis:
My only (slight) concern is that I believe this wasn't thread safe in 5.9

I don't know if it's still the case - or if it matters much at all. Would need to verify that

GNMoseke:
Yeah I saw that in the issue - I can verify in 5.9, will draft this guy for now.

GNMoseke:
At least as of `5.9.1`, the implementation of `ProcessInfo.processInfo.environment` is [the same as what this PR replaces](https://github.com/swiftlang/swift-corelibs-foundation/blob/release/5.9.1/Sources/Foundation/ProcessInfo.swift#L40) -  so there should be no difference for 5.9

The [underlying access](https://github.com/swiftlang/swift-corelibs-foundation/blob/465203cbe768e7c90763d3896aa53a6324ee2745/Sources/CoreFoundation/CFPlatform.c#L1876) is indeed a char** as @adam-fowler commented in the original issue - I am sorry to say that my low-level cpp knowledge isn't super strong, so this appears thread safe to me, but take that with a grain of salt

Joannis (approved):
Hmm, you're right. I just looked it up: `environ` is not thread safe. Happy to accept this change, but we'll need synchronisation it seems.

Sources/Hummingbird/Environment.swift
```diff
@@ -111,22 +112,8 @@ public struct Environment: Sendable, Decodable, ExpressibleByDictionaryLiteral {
...
-        while let entry = envp.advanced(by: idx).pointee {
-            let entry = String(cString: entry)
-            if let i = entry.firstIndex(of: equalSign) {
-                let key = String(entry.prefix(upTo: i))
-                let value = String(entry.suffix(from: i).dropFirst())
-                values[key.lowercased()] = value
-            }
-            idx += 1
+        ProcessInfo.processInfo.environment.keys.forEach { key in
+            values[key.lowercased()] = ProcessInfo.processInfo.environment[key]
```
Joannis, line 116:
```suggestion
        for key in ProcessInfo.processInfo.environment.keys {
            values[key.lowercased()] = ProcessInfo.processInfo.environment[key]
```

Joannis (commented):
Hmm, you're right. I just looked it up: `environ` is not thread safe. Happy to accept this change, but we'll need synchronisation it seems.

Sources/Hummingbird/Environment.swift
GNMoseke: Fair enough, fixed.

Sources/Hummingbird/Environment.swift
```diff
@@ -111,22 +112,8 @@ public struct Environment: Sendable, Decodable, ExpressibleByDictionaryLiteral {
...
-
-        while let entry = envp.advanced(by: idx).pointee {
-            let entry = String(cString: entry)
-            if let i = entry.firstIndex(of: equalSign) {
-                let key = String(entry.prefix(upTo: i))
-                let value = String(entry.suffix(from: i).dropFirst())
-                values[key.lowercased()] = value
-            }
-            idx += 1
+        for key in ProcessInfo.processInfo.environment.keys {
```
adam-fowler, line 115: You could do this as 
```swift
for item in ProcessInfo.processing.environment {
```
And then use `item.key` and `item.value` to avoid the additional dictionary look up.

GNMoseke: Sure, good call.

adam-fowler (dismissed):
Given this is replacing like for like and in Swift 6 processInfo.environment is thread safe I'm happy to accept this PR. We can discuss any synchronisation support for 5.x in another PR. Plus we should verify it is an issue.

GNMoseke:
Alright fellas sorry it took me so long to get back to this - as mentioned in #574 there's some thread safety issues here.

I **think** what was happening is that `ProcessInfo.processInfo`, as a static variable, along with `Environment.getEnvironment` as a static method were doing something funky when all tests were run in a single process (from a normal `swift test` invocation). About 1/5 times there I could reproduce the failure, including on a 6.0.1 toolchain.

I **could not** reproduce the failure running `swift test --parallel`, which is what leads me to believe that this only pops up in the case where all the tests are being started from the same process.

As a band-aid I renamed the env var that `testSetForAllEnvironments` uses, which feels like a bit of a hacky solution but does resolve the test failure.

TL;DR I don't think this is actually an issue in a real running hummingbird instance since the server is started as a single process and the environment is being properly mutated - just a race in tests.

adam-fowler (approved):

Joannis (approved):

PR #576, opened 2024-09-30
Clarify what the RequestContext is for

Joannis:


adam-fowler (approved):
One minor nit can we reduce the line length a little, but otherwise good.

Issue #565, opened 2024-09-27
Result builder `Body` style syntax.


connor-ricks:
Copying this issue over from the example repo to move it to the more appropriate location and closing the example issue.

Hello,

I was experimenting with Hummingbird today and saw the HummingbirdRouter target was added for result builder style syntax.

While I managed to use the result builder syntax to get some routes created, I was wondering if at this point in time, it was possible to make the content feel more composable.

Most of the examples build route collections using a property called endpoints but then when they add them, they explicitly call .endpoints.

I was hoping to achieve a syntax more like this...

```swift
let router = RouterBuilder(context: BasicRouterRequestContext.self) {
    Auth()
}

struct Auth: MiddlewareProtocol {
    var body: some MiddlewareProtocol {
        SSO()
        Get(...)
        Post(...)
    }
}

structure SSO: MiddlewareProtocol {
    var body: some MiddlewareProtocol {
        ...
    }
}
```

Just some rough syntax, but hopefully you get the idea.

Is this possible at the moment? Perusing the implementation, it seems maybe not quite, but perhaps it is pretty close.

I achieved a similar syntax with Vapor. You can take a peek [here](https://github.com/connor-ricks/swift-vapor-route-builder).

connor-ricks:
I managed to get this to compile this morning... Caching is still a potential issue. I'm wondering if this is a worthwhile addition to the library or if I should simply just use this in my own project?

```swift
protocol Controller: RouterMiddleware {
    associatedtype Body: RouterMiddleware
    var body: Body { get }
}

extension Controller where Body.Input == Input, Body.Context == Context, Body.Output == Output {
    func handle(_ input: Input, context: Context, next: (Input, Context) async throws -> Output) async throws -> Output {
        try await body.handle(input, context: context, next: next)
    }
}

struct UserController: Controller {
    typealias Context = BasicRouterRequestContext

    var body: some RouterMiddleware<Context> {
        RouteGroup("user") {
            Get { _,_ in
                "User"
            }
        }
    }
}

func foo() {
    let router = RouterBuilder {
        UserController()
    }

    let app = Application(responder: router)
}
```

adam-fowler:
It is a worthwhile addition. I think we'd want to integrate it into the result builder first though so the result builder is only evaluated at build time and not every time we call handle().

connor-ricks:
Thanks for the input. I might try and play around with this a little bit this weekend. Go enjoy your vacation!

[...]

PR #577, opened 2024-10-01
Add route controller for better composability. (#565)

connor-ricks:
### Description

This PR proposes the concept of a `RouterController ` which aims to improve the ergonomics of the `MiddlewareFixedTypeBuilder` result builder style syntax.

Conforming to `RouterController` should allow you to create composable routes in a more `SwiftUI` result builder style.

### Expectations
Here is a list of things that I would expect from the full result builder style syntax.

1. It shouldn't be necessary to call `body` on `RouterController`.
2. It should be possible for a `RouterController` `body` to be composed of `MiddlewareProtocol`, `RouterMiddleware` and other `RouterController` definitions using the appropriate result builder.
3. It should be possible to define a high order `RouterController` that takes a generic content result builder function. This is similar to how things like `VStack` and `Group` work in SwiftUI. This unlocks the ability to make reusable wrappers around nested content. Imagine a `Localized` controller that wraps some arbitrary content inside a specific path and adds a particular middleware.

### Struggles
Things I'm struggling with in the implementation...

1. ~The nuance between `@MiddlewareFixedTypeBuilder` and `@RouteBuilder` is a little lost on me, and I'm unsure which of the result builders `RouteController` should be using.~
2. ~Should `RouteController` itself conform to `RouterMiddleware` or `MiddlewareProtocol` or should additional builder expressions be added to support the varying types. In SwiftUI, the `View` protocol defines a `Body` that itself is a `View`. Working through this problem, separating out the `RouteController` `body` as a different requirement to itself has proved complex. I'm unsure of the pros/cons of each approach.~
3. ~Piping the `Context` through. in such a way that doesn't require the `body` definitions to explicitly define generic types or requiring complex where clauses, has proven quite complex.~

With all this said, I'm hoping for a more complete conversation/assistance with implementation architecture thoguhts before fiddling around more with implementation, I think there are quite a few more nuances/catches than I had anticipated when making my initial attempt. It's been fun to think through this more, but I don't presume to have the full context of where exactly the aim is to take hummingbird as a project.

Looking forward to your input @adam-fowler and @Joannis. 😄 Thanks for the contribution to the conversation so far.

### DIFF
--- a/Sources/Hummingbird/Middleware/MiddlewareModule/MiddlewareStack.swift
+++ b/Sources/Hummingbird/Middleware/MiddlewareModule/MiddlewareStack.swift
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+// MARK: - Middleware2
+
 public struct _Middleware2<M0: MiddlewareProtocol, M1: MiddlewareProtocol>: MiddlewareProtocol where M0.Input == M1.Input, M0.Context == M1.Context, M0.Output == M1.Output {
     public typealias Input = M0.Input
     public typealias Output = M0.Output
@@ -33,6 +35,10 @@ public struct _Middleware2<M0: MiddlewareProtocol, M1: MiddlewareProtocol>: Midd
         }
     }
 }
+
+extension _Middleware2: RouterMiddleware where M0.Input == Request, M0.Output == Response {}
+
+// MARK: - MiddlewareFixedTypeBuilder
 
 /// Middleware stack result builder
 ///
--- a/Sources/HummingbirdRouter/RouterController.swift
+++ b/Sources/HummingbirdRouter/RouterController.swift
@@ -1,0 +1,47 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2023-2024 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Hummingbird
+
+// MARK: - RouterController
+
+/// A type that represents part of your app's middleware and routes
+///
+/// You create custom controllers by declaring types that conform to the `RouterController`
+/// protocol. Implement the required ``RouterController/body-swift.property`` computed
+/// property to provide the content for your custom controller.
+///
+///     struct MyController: RouterController {
+///         typealias Context = BasicRouterRequestContext
+///
+///         var body: some RouterMiddleware<Context> {
+///             Get("foo") { _,_ in "foo" }
+///         }
+///    }
+///
+/// Assemble the controller's body by combining one or more of the built-in controllers or middleware.
+/// provided by Hummingbird, plus other custom controllers that you define, into a hierarchy of controllers.
+public protocol RouterController<Context> {
+    associatedtype Context
+    associatedtype Body: RouterMiddleware<Context>
+    @MiddlewareFixedTypeBuilder<Request, Response, Context> var body: Body { get }
+}
+
+// MARK: MiddlewareFixedTypeBuilder + RouterController Builders
+
+extension MiddlewareFixedTypeBuilder {
+    public static func buildExpression<C0: RouterController>(_ c0: C0) -> C0.Body where C0.Body.Input == Input, C0.Body.Output == Output, C0.Body.Context == Context {
+        return c0.body
+    }
+}
--- a/Tests/HummingbirdRouterTests/ControllerTests.swift
+++ b/Tests/HummingbirdRouterTests/ControllerTests.swift
@@ -1,0 +1,152 @@
+//===----------------------------------------------------------------------===//
+//
+// This source file is part of the Hummingbird server framework project
+//
+// Copyright (c) 2021-2023 the Hummingbird authors
+// Licensed under Apache License v2.0
+//
+// See LICENSE.txt for license information
+// See hummingbird/CONTRIBUTORS.txt for the list of Hummingbird authors
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import Hummingbird
+import HummingbirdRouter
+import HummingbirdTesting
+import XCTest
+
+final class ControllerTests: XCTestCase {
+    func testRouterControllerWithSingleRoute() async throws {
+        struct TestController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            var body: some RouterMiddleware<Context> {
+                Get("foo") { _, _ in "foo" }
+            }
+        }
+
+        let router = RouterBuilder(context: BasicRouterRequestContext.self) {
+            TestController()
+        }
+
+        let app = Application(responder: router)
+        try await app.test(.router) { client in
+            try await client.execute(uri: "/foo", method: .get) {
+                XCTAssertEqual(String(buffer: $0.body), "foo")
+            }
+        }
+    }
+
+    func testRouterControllerWithMultipleRoutes() async throws {
+        struct TestController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            var body: some RouterMiddleware<Context> {
+                Get("foo") { _, _ in "foo" }
+                Get("bar") { _, _ in "bar" }
+            }
+        }
+
+        let router = RouterBuilder(context: BasicRouterRequestContext.self) {
+            TestController()
+        }
+
+        let app = Application(responder: router)
+        try await app.test(.router) { client in
+            try await client.execute(uri: "/foo", method: .get) {
+                XCTAssertEqual(String(buffer: $0.body), "foo")
+            }
+
+            try await client.execute(uri: "/bar", method: .get) {
+                XCTAssertEqual(String(buffer: $0.body), "bar")
+            }
+        }
+    }
+
+    func testRouterControllerWithGenericChildren() async throws {
+        struct ChildController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            let name: String
+            var body: some RouterMiddleware<Context> {
+                Get("child_\(self.name)") { _, _ in "child_\(self.name)" }
+            }
+        }
+
+        struct ParentController<Context: RouterRequestContext, Child: RouterMiddleware>: RouterController where Child.Context == Context {
+            var child: Child
+
+            init(@MiddlewareFixedTypeBuilder<Request, Response, Context> _ child: () -> Child) {
+                self.child = child()
+            }
+
+            var body: some RouterMiddleware<Context> {
+                RouteGroup("parent") {
+                    self.child
+                }
+            }
+        }
+
+        let router = RouterBuilder(context: BasicRouterRequestContext.self) {
+            ParentController {
+                Get("child_a") { _, _ in "child_a" }
+                ChildController(name: "b")
+                ChildController(name: "c")
+                Get("child_d") { _, _ in "child_d" }
+            }
+        }
+
+        let app = Application(responder: router)
+        try await app.test(.router) { client in
+            for letter in "abcd" {
+                try await client.execute(uri: "/parent/child_\(letter)", method: .get) {
+                    XCTAssertEqual(String(buffer: $0.body), "child_\(letter)")
+                }
+            }
+        }
+    }
+
+    func testRouterControllerWithMiddleware() async throws {
+        struct TestMiddleware<Context: RequestContext>: RouterMiddleware {
+            func handle(_ request: Request, context: Context, next: (Request, Context) async throws -> Response) async throws -> Response {
+                var response = try await next(request, context)
+                response.headers[.middleware] = "TestMiddleware"
+                return response
+            }
+        }
+
+        struct ChildController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            var body: some RouterMiddleware<Context> {
+                Get("foo") { _, _ in "foo" }
+            }
+        }
+
+        struct ParentController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            var body: some RouterMiddleware<Context> {
+                RouteGroup("parent") {
+                    TestMiddleware()
+                    ChildController()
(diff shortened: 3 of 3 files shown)


adam-fowler:
> 1. The nuance between `@MiddlewareFixedTypeBuilder` and `@RouteBuilder` is a little lost on me, and I'm unsure which of the result builders `RouteController` should be using.

I think you should ignore `@RouteBuilder` for this PR. 

> 2. Should `RouteController` itself conform to `RouterMiddleware` or `MiddlewareProtocol` or should additional builder expressions be added to support the varying types. In SwiftUI, the `View` protocol defines a `Body` that itself is a `View`. Working through this problem, separating out the `RouteController` `body` as a different requirement to itself has proved complex. I'm unsure of the pros/cons of each approach.

RouteController should not conform to MiddlewareProtocol. In SwiftUI there is a requirement for a View to have a body. This requirement does not exist for MiddlewareProtocol. 

> 3. Piping the `Context` through. in such a way that doesn't require the `body` definitions to explicitly define generic types or requiring complex where clauses, has proven quite complex.

Without having access to a computer I cant really help you here. My internal swift compiler falls apart when it comes to generics. You’ll have to wait until I get back from holiday.

Sources/Hummingbird/Middleware/MiddlewareController.swift
```diff
@@ -0,0 +1,189 @@
...
+
+    public static func buildEither<C0: MiddlewareController, M0: MiddlewareProtocol>(
+        second content: M0
+    ) -> _ConditionalMiddleware<C0, M0> {
+        _ConditionalMiddleware.second(content)
+    }
+}
+
+#warning("This is a playground of concrete controllers to help validate the protocol during development.")
+#warning("Once the implementation is complete, delete this and migrate the work to unit tests.")
```
connor-ricks, line 95: - [ ] Delete playgrounds once we are happy with controller definitions.

Sources/Hummingbird/Router/RouterController.swift
```diff
@@ -0,0 +1,119 @@
...
+///    }
+///
+/// Assemble the controller's body by combining one or more of the built-in controllers or middleware.
+/// provided by Hummingbird, plus other custom controllers that you define, into a hierarchy of controllers.
+public protocol RouterController<Context>: MiddlewareController where Body: RouterMiddleware<Context> {
+    associatedtype Context
+}
+
+#warning("This is a playground of concrete controllers to help validate the protocol during development.")
+#warning("Once the implementation is complete, delete this and migrate the work to unit tests.")
```
connor-ricks, line 39: - [ ] Delete playgrounds once we are happy with controller definitions.

Sources/Hummingbird/Router/RouterController.swift
```diff
@@ -0,0 +1,119 @@
...
+private struct ControllerWithRouterControllerChild<Child: RouterController>: RouterController {
+    typealias Context = Child.Context
+
+    @MiddlewareFixedTypeBuilder<Input, Output, Context> var child: () -> Child
+
+    var body: some RouterMiddleware<Context> {
+        child()
+        #warning("This IS NOT able to infer Context or the correct result builder function.")
+//        LogRequestsMiddleware(.info)
+//        MetricsMiddleware()
```
connor-ricks, line 85: - [ ] Struggling to get this compiling

Sources/Hummingbird/Middleware/MiddlewareController.swift
```diff
@@ -0,0 +1,189 @@
...
+    typealias Output = Child.Output
+    typealias Context = Child.Context
+
+    @MiddlewareFixedTypeBuilder<Input, Output, Context> var child: () -> Child
+
+    var body: some MiddlewareProtocol<Input, Output, Context> {
+        child()
+        #warning("This IS NOT able to infer Context or the correct result builder function.")
+//        LogRequestsMiddleware(.info)
+//        MetricsMiddleware()
```
connor-ricks, line 169: - [ ] Struggling to get this compiling

Sources/Hummingbird/Middleware/MiddlewareController.swift
```diff
@@ -0,0 +1,189 @@
...
```
connor-ricks, line 149: - [ ] Struggling to get this compiling

Sources/Hummingbird/Middleware/MiddlewareModule/MiddlewareStack.swift
```diff
@@ -34,6 +36,47 @@ public struct _Middleware2<M0: MiddlewareProtocol, M1: MiddlewareProtocol>: Midd
     }
 }
 
+extension _Middleware2: RouterMiddleware where M0.Input == Request, M0.Output == Response {}
+
+// MARK: - ConditionalMiddleware
+
+public enum _ConditionalMiddleware<M0: MiddlewareProtocol, M1: MiddlewareProtocol>: MiddlewareProtocol where M0.Input == M1.Input, M0.Context == M1.Context, M0.Output == M1.Output {
```
connor-ricks, line 43: - [ ] With the addition of `_ConditionalMiddleware` and `_EmptyMiddleware`, this file is getting a bit bulky. It might be worthwhile moving all these "public private" middleware unto their own files inside a subfolder.

adam-fowler:
This PR seems to have gained a lot more code. Can we roll it back to the original idea of a RouteController, without all the optional, eitherOr options. These aren't supported for MiddlewareProtocol so adding them for MiddlewareController doesn't make sense. 

If you feel they are needed we can discuss that in a separate PR.

connor-ricks:
> This PR seems to have gained a lot more code. Can we roll it back to the original idea of a RouteController, without all the optional, eitherOr options. These aren't supported for MiddlewareProtocol so adding them for MiddlewareController doesn't make sense.
> 
> If you feel they are needed we can discuss that in a separate PR.

Hey @adam-fowler ,

Yea, I can extract the additional result builder functionality into a separate PR.

Fwiw, a decent amount of this code is just "example" types that I expect to delete once I can see the compiler supports all the various "playground" types.

I'm a curious on your thoughts about having added both `RouteController` and `MiddlewareController`. It seemed to follow the patterns you had outlined for `MiddlewareProtocol` and `RouterMiddleware` a bit better.

connor-ricks:
I've removed the additional result builder functionality.

adam-fowler:
Hi Connor,
I've eventually got access to a laptop and I'm gonna have to apologise. I don't think the MiddlewareController adds anything in the end. I think we should probably just limit everything to RouterController. This is what comes from trying to review PRs on a phone.
With that everything pretty much reduces to the following, which isn't far off what you originally posted 
```swift
public protocol RouterController<Context> {
    associatedtype Context
    associatedtype Body: RouterMiddleware<Context>
    @MiddlewareFixedTypeBuilder<Request, Response, Context> var body: Body { get }
}

extension MiddlewareFixedTypeBuilder {
    public static func buildExpression<C0: RouterController>(_ c0: C0) -> C0.Body where C0.Context == Context {
        return c0.body
    }

    public static func buildBlock<C0: RouterController>(_ c0: C0) -> C0.Body {
        return c0.body
    }

    public static func buildPartialBlock<C0: RouterController>(first: C0) -> C0.Body {
        first.body
    }

    public static func buildPartialBlock<M0: MiddlewareProtocol, C0: RouterController>(
        accumulated m0: M0,
        next c0: C0
    ) -> _Middleware2<M0, C0.Body> where M0.Input == Request, M0.Output == Response, M0.Context == C0.Context {
        _Middleware2(m0, c0.body)
    }
}
```
Its a little frustrating we have to conform `_Middleware2` to `RouterMiddleware` but I don't see any way around that.

connor-ricks:
Hey @adam-fowler, no worries at all, and I appreciate you taking the time to provide your input and context. I hope your vacation was great!

I agree on the value of `MiddlewareController` being minimal. I only added it to follow what I viewed as the pattern of the codebase, but I agree it's not really brining more or less to the table here. I've updated the PR to remove it.

connor-ricks:
@adam-fowler I did add a unit test here that I was struggling to get compiling, and I wonder if you might have some insight to the solution.

Checkout `testRouterControllerWithGenericChild` which has a `ParentController` that takes a generic `Child` `RouterController`.

Tests/HummingbirdRouterTests/ControllerTests.swift
```diff
@@ -0,0 +1,78 @@
...
+//
+//===----------------------------------------------------------------------===//
+
+import Hummingbird
+import HummingbirdRouter
+import HummingbirdTesting
+import XCTest
+
+final class ControllerTests: XCTestCase {
+    func testRouterController() async throws {
```
adam-fowler, line 21: We should add tests that test all of the functions in `MiddlewareFixedTypeBuilder` so we need a test with a single route inside a controller.

adam-fowler (changes requested):
It all looks good now. A single issue to resolve and a solution to your child controller

Tests/HummingbirdRouterTests/ControllerTests.swift
```diff
@@ -0,0 +1,78 @@
...
+
+    func testRouterControllerWithGenericChild() async throws {
+        struct ChildController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            var body: some RouterMiddleware<Context> {
+                Get("child") { _,_ in "child" }
+            }
+        }
+
+        struct ParentController<Child: RouterController>: RouterController where Child.Context: RouterRequestContext {
```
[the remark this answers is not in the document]
connor-ricks: Ahh 👍 Yes, that totally makes sense. I got lost in the generic soup.

Tests/HummingbirdRouterTests/ControllerTests.swift
connor-ricks: I added a few tests, and actually removed some of the added result builders.

I think we don't actually need any of the additional result builders except for `buildExpression` Looking at the code coverage, basically all the implementations flow through the `RouterController` `buildExpression` and then back into the normal result builders defined in `Hummingbird`

```
public static func buildExpression<C0: RouterController>(_ c0: C0) -> C0.Body where C0.Body.Input == Input, C0.Body.Output == Output, C0.Body.Context == Context {
    return c0.body
}
```

connor-ricks: I wish that parameter packs supported Same-element requirements, it would allow us to replace `_Middleware2` with something like this

```
public struct TupleMiddleware<Input, Output, Context, each M: Handler<Input, Output, Context>>: MiddlewareProtocol {
    public func handle(_ input: Input, context: Context) -> Output {
        ...
    }
}
```

this would greatly reduce the type signature craziness when looking at errors.

i.e stuff like this 

`Generic struct 'ParentController' requires that '_Middleware2<_Middleware2<_Middleware2<Route<_RouteHandlerClosure<String, BasicRouterRequestContext>, BasicRouterRequestContext>, some RouterMiddleware<BasicRouterRequestContext>>, some RouterMiddleware<BasicRouterRequestContext>>, Route<_RouteHandlerClosure<String, BasicRouterRequestContext>, BasicRouterRequestContext>>' ....etc......`

connor-ricks: I'm assuming you'd be opposed to the old-fashioned pre-parameter-pack `_Middleware2`, `_Middleware3`, `_Middleware4` etc...

adam-fowler: I'm not going to change it now. I can't see how you'd do it without breaking changes. 

The MiddlewareProtocol is closely mapped to a generic middleware being developed by members of SSWG (unavailable at the moment). It's just taking a long time because key members are very busy.

adam-fowler: > I think we don't actually need any of the additional result builders except for buildExpression

Ok if you can show via tests that all expected cases are covered by `buildExpression` then we should remove the others.

connor-ricks: I think the additional tests should cover the basis. I'll do another pass in an hour or so, just popping out for an early dinner.

connor-ricks: I added one more test to validate middleware compiles correctly and executes.

Sources/HummingbirdRouter/RouterController.swift
```diff
@@ -0,0 +1,63 @@
...
+
+
+// MARK: MiddlewareFixedTypeBuilder + RouterController Builders
+
+extension MiddlewareFixedTypeBuilder {
+    public static func buildExpression<C0: RouterController>(_ c0: C0) -> C0.Body where C0.Body.Input == Input, C0.Body.Output == Output, C0.Body.Context == Context {
+        return c0.body
+    }
+
+//    public static func buildBlock<C0: RouterController>(_ c0: C0) -> C0.Body {
```
adam-fowler, line 49: Let's delete the commented out code.

Tests/HummingbirdRouterTests/ControllerTests.swift
```diff
@@ -0,0 +1,148 @@
...
+        struct ChildController: RouterController {
+            typealias Context = BasicRouterRequestContext
+            let name: String
+            var body: some RouterMiddleware<Context> {
+                Get("child_\(name)") { _,_ in "child_\(name)" }
+            }
+        }
+
+        struct ParentController<Context: RouterRequestContext, Child: RouterMiddleware>: RouterController where Child.Context == Context {
+            @MiddlewareFixedTypeBuilder<Request, Response, Context> var child: () -> Child
```
adam-fowler, line 76: Swift 5.9 doesn't seem to like this. Perhaps instead of storing the closure you store the result of the closure it might work.

adam-fowler (changes requested):

Tests/HummingbirdRouterTests/ControllerTests.swift
connor-ricks: Ahh, yea I'm running on latest Xcode with 6.0. I'll check compilation on 5.9.

connor-ricks: I didn't have any issues compiling and running using this command...

`swift test -Xswiftc -swift-version -Xswiftc 5`

adam-fowler: It is Swift 5.9 compiler that is the issue, not the Swift 6 compiler in Swift 5 mode.

Joannis (approved):
Looks great to me!

adam-fowler (approved):
Looks good

PR #582, opened 2024-10-09
Use new discord invite in README

adam-fowler:
Don't merge this. I've found the new invite seems to have expired


adam-fowler:
Don't merge this. I've found the new invite seems to have expired

Joannis (approved):

Issue #578, opened 2024-10-02
Onboarding wizard feedback


trevor-e:
Hey, been following along and am excited for this project. I gave it a spin on and had some feedback for the onboarding/new project process, figured it made sense to group it all here.

1. Compared to other frameworks, it was a bit unexpected that we clone a template repo, run a command, and then it spits out a new sub-repo. It took me a second to realize what was happening. I figure there are eventual plans to do this fully within a CLI, just wanted to point out my confusion.
2. `Enter your package name:` package name is such an overloaded term I thought it was asking for a bundle ID. Maybe it could be phrased as "Swift Package name" to be unambiguous. 
3. Some of the console output seemed distracting, like it just printed out `~/dev/hummingbird-template` a few times.
4. It would be great if the console output used colors to separate some of the information.
5. `Enter the folder created and run 'swift run' to build and run your server.` coupled with my confusion earlier about how this creates a sub-directory with the scaffolded project, I was thinking "what folder"? Could it explicitly tell me which folder to open?
6. `Then open 'localhost:8080' on your web browser.` I don't think the CLI needs to say this since `swift run` already outputs this too.
7. When I run `swift run` it actually outputs `[HummingbirdCore] Server started and listening on 127.0.0.1:8080`, it would be nice if it said `localhost:8080` for consistency. VSCode also seems to stumble trying to CMD+click opening `127.0.0.1:8080` into the browser so it would probably work better with localhost like other frameworks use.

Other than that everything worked! Excited to follow along.

adam-fowler:
> Hey, been following along and am excited for this project. I gave it a spin on and had some feedback for the onboarding/new project process, figured it made sense to group it all here.
> 
> 
> 
> 1. Compared to other frameworks, it was a bit unexpected that we clone a template repo, run a command, and then it spits out a new sub-repo. It took me a second to realize what was happening. I figure there are eventual plans to do this fully within a CLI, just wanted to point out my confusion.
> 
If you run the `configure.sh` command without any parameters it should edit the template in place. I have to play with this a little to get it right.

> 2. `Enter your package name:` package name is such an overloaded term I thought it was asking for a bundle ID. Maybe it could be phrased as "Swift Package name" to be unambiguous. 
> 
Fair point

> 3. Some of the console output seemed distracting, like it just printed out `~/dev/hummingbird-template` a few times.
> 
Yes this could be cleaned up

> 4. It would be great if the console output used colors to separate some of the information.
> 
👍 
[...]

PR #585, opened 2024-10-13
Print the server's scheme when logging that it's reachable

Joannis:


Joannis:
Not sure if we can make this change - but I think it would be nice to log the scheme as described in #578. We can also assume HTTP (not HTTPS) here, but I'd prefer logging it correctly.

Joannis:
We can also consider detecting if the IP is `::` or `0.0.0.0` so that we can print more accurate info

adam-fowler (commented):
Seems a bit of overkill just to print `http` or `https`.
Could you not just check the first channel handler is `NIOSSLServerHandler`

Joannis:
Unfortunately no, as TransportServices does not have a handler, and these handlers are only applicable for child channels of the server

adam-fowler (commented):
I'm not sure what the point of this is? Why do you want to print 127.0.0.1 when the bind address is 0.0.0.0

Joannis:
@adam-fowler my idea was that I want to (eventually) make it explicit how to reach it from various devices. Should've explained that in the PR.

So `0.0.0.0` should also show your LAN IP

adam-fowler:
@Joannis But bind address 0.0.0.0 is different from 127.0.0.1, ie one accepts connections from any address while the other accepts only connections from the localhost. We shouldn't be pretending they are the same.

Maybe if the bind address is set to 0.0.0.0 then we can add an additional log that says it is accepting connections from anywhere

Joannis:
I understand what `0.0.0.0` does, but it's not common knowledge. By adding logs about the available addresses to connect to, I'm hoping to address that knowledge gap so that people understand how this works.

Joannis:
An excellent example that I've had people ask time and time again:
> How do I connect to my backend from an iOS app? `127.0.0.1` doesn't work.

adam-fowler:
You need an additional log item then, because the people who do know what is going on are going to be a little confused when there server logs says the server is listening on 127.0.0.1 but accepting connections from anywhere. 

> An excellent example that I've had people ask time and time again:
> 
> > How do I connect to my backend from an iOS app? `127.0.0.1` doesn't work.

Also can we make it clear which situation we are trying to fix here, as your change doesn't fix this.

PR #590, opened 2024-10-14
Add URI.decodeQuery

adam-fowler:
Currently to decode the query string of a URL it is up to the user to create a `URLEncodedFormDecoder`. This PR adds a helper function `Request.decodeQuery` that creates the decoder and deals with any errors it might create when calling decode


Sources/Hummingbird/Server/Request.swift
```diff
@@ -55,6 +55,34 @@ extension Request {
             throw error
         }
     }
+
+    /// Decode request query using ``URLEncodedFormDecoder``.
+    /// - Parameters
+    ///   - type: Type you want to decode to
+    ///   - context: Request context
+    public func decodeQuery<Type: Decodable>(as type: Type.Type, context: some RequestContext) throws -> Type {
```
Joannis, line 63: I think it's a _little_ awkward to discover this API. Can we possibly add it on `URI`?

adam-fowler: Moved to URI

Sources/Hummingbird/Server/URI+decodeQuery.swift
```diff
@@ -0,0 +1,45 @@
...
+//===----------------------------------------------------------------------===//
+
+import HummingbirdCore
+
+extension URI {
+    /// Decode request query using ``URLEncodedFormDecoder``.
+    /// - Parameters
+    ///   - type: Type you want to decode to
+    ///   - context: Request context
+    public func decodeQuery<Type: Decodable>(as type: Type.Type, context: some RequestContext) throws -> Type {
```
Joannis, line 22: Idea: We could add add type inferrence here:
```suggestion
    public func decodeQuery<Type: Decodable>(as type: Type.Type = Type.self, context: some RequestContext) throws -> Type {
```

Joannis (approved):
LGTM, one small idea

PR #593, opened 2024-10-14
Add permissions pullrequest:write to benchmark gh action

adam-fowler:


Joannis:
What's this needed for?

adam-fowler:
> What's this needed for?

Writing the comment to the PR. Although I read later this won't work. I need to make some more changes to get that to work.

PR #596, opened 2024-10-23
Throwing RequestContext init from parent RequestContext

adam-fowler:


Sources/Hummingbird/Router/RouterMethods.swift
```diff
@@ -83,6 +83,32 @@ extension RouterMethods {
         )
     }
 
+    /// Return a group inside the current group that transforms the ``RequestContext``
+    ///
+    /// For the transform to work the `Source` of the transformed `RequestContext` needs
+    /// to be the original `RequestContext` eg
```
Joannis, line 89: Document what happens if the child initialization fails

Joannis (commented):
I'm really happy with this PR's shape

Joannis (approved):

PR #598, opened 2024-10-25
Remove @discardableResult from RouterMethods.group functions

adam-fowler:
It doesn't make sense to create a router group and do nothing with it and this shouldn't be encouraged
This is a breaking change, but any code relying on this is already broken.


Joannis (approved):
100% in favour, I thought we made this PR already 😅

PR #609, opened 2024-11-15
Renamed SameSite.secure to .strict

adam-fowler:
Realised that `secure` is not a valid value for the SameSite parameter


thoven87:
Should a check be introduced where if same site is set to none secure must also set? That's how the standard defines Samesite? Or maybe, secure would be set automatically if same site is none.

adam-fowler:
I did think about that. I don't think setting secure automatically is a great idea as it will confuse. We are better failing building the cookie or we could add an assert.

Joannis (approved):

PR #610, opened 2024-11-16
Assert on cookies with SameSite=None and no secure attribute

adam-fowler:


Sources/Hummingbird/HTTP/Cookie.swift
```diff
@@ -102,6 +102,7 @@ public struct Cookie: Sendable, CustomStringConvertible {
         httpOnly: Bool = true,
         sameSite: SameSite
     ) {
+        assert(!(secure == false && sameSite == .none), "Cookies with SameSite set to None require the Secure attribute to be set")
```
Joannis, line 105: I'm not sure if we should crash on these (even on DEBUG) - I know it's incorrect code that would lead to this. I'm not convinced that _this type_ or programming error should be something we - as a framework - should trigger on.

adam-fowler: How do we report this then? print an error to the log? If we don't point this out to the user they'll get unexplained things happening. Calling the Cookie initialiser with SameSite set to none and secure set to false is programmer error and should be indicated.

Joannis (approved):
I'm fine accepting this as is, but would still prefer getting extra input from folks around the ecosystem.

PR #612, opened 2024-11-16
Split HTTP2 Channel out from upgrade channel

adam-fowler:
This splits out the HTTP2Channel parts of the HTTP2UpgradeChannel into its own ServerChildChannel so it can be used by other ServerChildChannels eg a gRPC HTTP2 negotiation.

### DIFF
--- a/Sources/HummingbirdHTTP2/HTTP2Channel.swift
+++ b/Sources/HummingbirdHTTP2/HTTP2Channel.swift
@@ -21,7 +21,7 @@ import NIOHTTPTypesHTTP2
 import NIOSSL
 
 /// Child channel for processing HTTP2
-public struct HTTP2Channel: ServerChildChannel {
+internal struct HTTP2Channel: ServerChildChannel {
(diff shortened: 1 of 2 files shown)


Sources/HummingbirdHTTP2/HTTP2Channel.swift
```diff
@@ -16,25 +16,19 @@ import HTTPTypes
...
-import NIOPosix
 import NIOSSL
-import NIOTLS
 
-/// Child channel for processing HTTP1 with the option of upgrading to HTTP2
-public struct HTTP2UpgradeChannel: HTTPChannelHandler {
-    typealias HTTP1ConnectionOutput = HTTP1Channel.Value
-    typealias HTTP2ConnectionOutput = NIOHTTP2Handler.AsyncStreamMultiplexer<HTTP2StreamChannel.Value>
+/// Child channel for processing HTTP2
+public struct HTTP2Channel: ServerChildChannel {
```
Joannis, line 24: Can we keep this `internal` until we have a real use case using this?

Joannis (approved):

Issue #621, opened 2024-12-02
Fix ApplicationRequestContextSource has non-sendable type


vamsii777:
The `CoreRequestContextStorage` struct is marked as Sendable but contains a non-sendable property from RequestContextSource. We need to add Sendable conformance to fix thread safety warnings.

Steps to reproduce:
1. Use `CoreRequestContextStorage` in concurrent context
2. Compiler warns about non-sendable type `ApplicationRequestContextSource`

```swift
// MARK: - Core Context Storage
struct CoreRequestContextStorage {
    let source: ApplicationRequestContextSource
    var parameters: Parameters
    var allocator: ByteBufferAllocator
    
    init(source: ApplicationRequestContextSource) {
        self.source = source
        self.parameters = .init()
        self.allocator = .init()
    }
}
```

Environment:
- OS: macOS 15.1.1 (24B91)
- Swift Version: 6.0

adam-fowler:
`CoreRequestContextStorage` doesn't store an instance of `ApplicationRequestContextSource` so not sure why it needs to set to Sendable.

vamsii777:
Sorry about that! My implementation wasn’t quite right. I double-checked the docs and fixed the issue.

PR #622, opened 2024-12-02
Mark RequestContextSource as Sendable

vamsii777:
This PR adds Sendable conformance to the `RequestContextSource` protocol and ensures that `ApplicationRequestContextSource` properties are Sendable. 

- #621


adam-fowler:
This is a breaking change

PR #624, opened 2024-12-03
Add computed variable Router.routes

adam-fowler:
Returns a list of all the endpoints added to the router

### DIFF
--- a/Sources/Hummingbird/Router/Router.swift
+++ b/Sources/Hummingbird/Router/Router.swift
@@ -131,20 +131,20 @@ public struct RouterOptions: OptionSet, Sendable {
 
 extension Router {
     /// Endpoint description
-    public struct EndpointDescription: CustomStringConvertible {
+    public struct RouteDescription: CustomStringConvertible {
         public let path: RouterPath
         public let method: HTTPRequest.Method
 
         public var description: String { "\(method): \(path)" }
     }
 
-    /// List of router endpoints
-    public var endpoints: [EndpointDescription] {
+    /// List of router routes
+    public var routes: [RouteDescription] {
         let trieValues = self.trie.root.values()
         return trieValues.flatMap { endpoint in
             endpoint.value.methods.keys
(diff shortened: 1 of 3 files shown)


Sources/Hummingbird/Router/Router.swift
```diff
@@ -128,3 +128,25 @@ public struct RouterOptions: OptionSet, Sendable {
...
+
+extension Router {
+    /// Route description
+    public struct RouteDescription: CustomStringConvertible {
+        /// Route path
+        public let path: RouterPath
+        /// Route method
+        public let method: HTTPRequest.Method
+
+        public var description: String { "\(method): \(path)" }
```
Joannis, line 140: Can we print it without colon? E.G.

```
GET /users/
POST /users/
```

```suggestion
        public var description: String { "\(method) \(path)" }
```

Sources/Hummingbird/Router/TrieRouter.swift
```diff
@@ -91,3 +91,20 @@ import HummingbirdCore
         }
     }
 }
+
+extension RouterPathTrieBuilder.Node {
+    /// Return list of paths and associated values in trie node
+    /// - Parameter prefix: Prefix for path
+    /// - Returns: Array of path values pairs
+    func values(prefix: RouterPath = "/") -> [(value: Value, path: RouterPath)] {
```
Joannis, line 99: I don't think we care about performance here, but for the sake of having mentioned it we should consider a syntax with `into values: inout [(value: Value, path: RouterPath)]`. This allows us to limit (re)allocs of arrays.

Joannis (approved):

Joannis:
Two nits

Sources/Hummingbird/Router/TrieRouter.swift
adam-fowler: For the sake of readability and given we don't care about performance here I'll keep it as is

PR #627, opened 2024-12-05
Cancel on inbound close

adam-fowler:
Add support for cancelling a request when the inbound is closed.
```swift
try await request.cancelOnInboundClose { request in
    try await doThingThatTakesALongTime()
}
```
There are limitations 
- `request.cancelOnInboundClose` consumes the request body
- It takes an escaping closure so cannot be used in a middleware to call the next middleware in the chain.

The way it works. It runs a task group with one task reading the request inbound stream, and the other running the closure. To avoid the request inbound stream being iterated twice a new version of the request is constructed with a new inbound stream that is being fed by the first of the tasks.

If the request inbound stream hits a second http head (ie the next request) then that is stored in the request iteration state attached to Request.

### DIFF
--- a/Sources/HummingbirdCore/Request/Request.swift
+++ b/Sources/HummingbirdCore/Request/Request.swift
@@ -13,7 +13,9 @@
 //===----------------------------------------------------------------------===//
 
 import HTTPTypes
+import NIOConcurrencyHelpers
 import NIOCore
+import NIOHTTPTypes
 
 /// Holds all the values required to process a request
 public struct Request: Sendable {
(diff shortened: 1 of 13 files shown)


Sources/HummingbirdCore/Request/Request.swift
```diff
@@ -60,10 +65,100 @@ public struct Request: Sendable {
         self.body = .init(buffer: byteBuffer)
         return byteBuffer
     }
+
+    public func cancelOnInboundClose<Value: Sendable>(_ process: @escaping @Sendable (Request) async throws -> Value) async throws -> Value {
```
Joannis, line 69: This should be part of the (Core)RequestContext instead

adam-fowler: `Server` needs access to this object and it knows nothing about RequestContexts so not sure how we can do this.

adam-fowler:
Need #629 to really make this useful

PR #628, opened 2024-12-05
Make RequestBody.Source Sendable

adam-fowler:
This is a requirement for the close on inbound. 
It is a breaking change, so I isolated it for review

### DIFF
--- a/Sources/HummingbirdCore/Request/RequestBody.swift
+++ b/Sources/HummingbirdCore/Request/RequestBody.swift
@@ -162,7 +162,7 @@ extension RequestBody {
     }
 
     /// A source used for driving a ``RequestBody`` stream.
-    public struct Source: Sendable {
+    public final class Source: Sendable {
         @usableFromInline
         let source: Producer.Source
         @usableFromInline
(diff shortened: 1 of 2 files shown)


Sources/HummingbirdCore/Request/RequestBody.swift
```diff
@@ -162,13 +162,13 @@ extension RequestBody {
     }
 
     /// A source used for driving a ``RequestBody`` stream.
-    public final class Source {
+    public struct Source: Sendable {
```
Joannis, line 165: The change from `class` -> `struct` is the breaking bit right? Can we keep this a class (for now)? It can still be Sendable, as all members are constants and Sendable

adam-fowler: yeah I've got the Source as a final class in the other PR so we can do that

adam-fowler: I reverted it to a final class and the benchmarks shows up the fact we have gained an extra allocation. This is a fairly rare path to take so isn't a huge issue. It is just one that the streaming benchmark hits all the time as we have to create streamed request.

PR #629, opened 2024-12-05
BREAKING CHANGE: Remove inout from writer in ResponseBody.init closure

adam-fowler:
Also use `sending` in swift 6 to pass ResponseBodyWriter into function so it can be used in a sending closure. 

This change is needed to make the `cancelOnInboundClosure` PR #627 useful. Without this you are not able to pass the ResponseBodyWriter into the `cancelOnInboundClosure` closure, without nasty unsafe wrapping of the type.

As this is a breaking change I have separated it out into its own PR. This change will most likely not affect anyone as I don't think many people are writing their own `ResponseBodyWriter`. There are a couple of minor changes in hummingbird-compression but it doesn't break it.

### DIFF
--- a/Sources/HummingbirdCore/Response/ResponseBody.swift
+++ b/Sources/HummingbirdCore/Response/ResponseBody.swift
@@ -17,8 +17,14 @@ import NIOCore
 
 /// Response body
 public struct ResponseBody: Sendable {
+    #if compiler(>=6.0)
+    public typealias WriteFunction = @Sendable (sending any ResponseBodyWriter) async throws -> Void
+    #else
+    public typealias WriteFunction = @Sendable (any ResponseBodyWriter) async throws -> Void
+    #endif
+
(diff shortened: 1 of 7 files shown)


adam-fowler:
Without this change the following code would not compile
```swift
return Response(
    status: .ok, 
    body: { writer in
        try await request.cancelOnInboundClosure { request in
            while true {
                try await Task.sleep(for: .seconds(1))
                try await writer.write(ByteBuffer(string: "tick")
            }
        }
    }
)
```
It would complain because it can't pass the writer into the `request.cancelOnInboundClosure` closure

adam-fowler:
> We should run the examples against this to spot how likely people are to run into this. I expect this to be fine though. I only know of one project that might need to handle this breaking change.

Run them all and as expected the ResponseBodyProcessing example needed a minor edit, but everything else is fine

PR #631, opened 2024-12-07
Cancel on inbound close (version 2)

adam-fowler:
Alternative to #627

This version manages to keep all the changes to `RequestBody`. But that does bring in an extra limitation (you can't set the request body to a new type before calling `consumeWithInboundCloseHandler` as we lose the connection to the underlying iterator.


Tests/HummingbirdCoreTests/TestUtils.swift
```diff
@@ -80,22 +80,38 @@ public func testServer<Value: Sendable>(
...
+            try await serviceGroup.run()
+        }
+        let port = await promise.wait()
         let client = TestClient(
             host: "localhost",
             port: port,
             configuration: clientConfiguration,
             eventLoopGroupProvider: .createNew
         )
+        print("Client connecting to port \(port)")
```
Joannis, line 110: This should be a log statement if we plan to keep it

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,142 @@
...
+    /// If the response finishes the connection will be closed.
+    ///
+    /// - Parameter operation: The actual operation to run
+    /// - Returns: Return value of operation
+    public func consumeWithCancelOnInboundClose<Value: Sendable>(
+        _ operation: sending @escaping (RequestBody) async throws -> Value
+    ) async throws -> Value {
+        let (barrier, source) = AsyncStream<Void>.makeStream()
+        return try await consumeWithInboundCloseHandler { body in
+            let unsafeOperation = UnsafeTransfer(operation)
```
Joannis, line 73: Can we achieve  the same thing by `sending` this operation in a capture syntax block? We're only using it in one place, so sending it would be fair

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,142 @@
...
+    ///
+    /// This function is designed for use with long running requests like server sent events. It assumes you
+    /// are not going to be using the request body after calling as it consumes the request body, it also assumes
+    /// you havent edited the request body prior to calling this function.
+    ///
+    /// If the response finishes the connection will be closed.
+    ///
+    /// - Parameter operation: The actual operation to run
+    /// - Returns: Return value of operation
+    public func consumeWithCancelOnInboundClose<Value: Sendable>(
```
Joannis, line 68:
```suggestion
    public func consumeWithCancellationOnInboundClose<Value: Sendable>(
```

or

```suggestion
    public func readWithCancellationOnInboundClose<Value: Sendable>(
```

I'd spell it out as `cancellation`, not `cancel`, since it's going to be in the act of being cancelled.

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,142 @@
...
+    /// you havent edited the request body prior to calling this function.
+    ///
+    /// If the response finishes the connection will be closed.
+    ///
+    /// - Parameters
+    ///   - operation: The actual operation
+    ///   = onInboundClose: handler invoked when inbound is closed
+    /// - Returns: Return value of operation
+    @available(macOS 15, iOS 18, tvOS 18, *)
+    public func consumeWithInboundCloseHandler<Value: Sendable>(
```
Joannis, line 36:
```suggestion
    public func readWithInboundCloseHandler<Value: Sendable>(
```

I think _read_ makes more sense, as `consume` could be interpreted differently

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,142 @@
...
+                group.cancelAll()
+                throw CancellationError()
+            }
+        } onInboundClosed: {
+            source.finish()
+        }
+    }
+
+    @available(macOS 15, iOS 18, tvOS 18, *)
+    func withInboundCloseHandler<Value: Sendable>(
```
Joannis, line 96:
```suggestion
    fileprivate func withInboundCloseHandler<Value: Sendable>(
```

Let's explicitly mark this as something for local use only

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,142 @@
...
+        return value
+    }
+
+    enum IterateResult {
+        case inboundClosed
+        case nextRequestReady
+    }
+
+    @available(macOS 15, iOS 18, tvOS 18, *)
+    func iterate(
```
Joannis, line 124:
```suggestion
    fileprivate func iterate(
```

Joannis (changes requested):
Some food for thought, but looks good in general

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
adam-fowler: Added cancellation. I'm not sure about the read vs consume. Read doesn't necessarily imply the request body will have been fully read once you leave the function.

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
adam-fowler: Nope capture syntax blocks don't help here. This is a real pain. I posted about it on the forums. As far as I can tell the capture in the `withThrowingTaskGroup` closure removes the sending attribute from the `operation` closure so I cannot then pass it to `TaskGroup.addTask`.

adam-fowler: You can see it reported here https://github.com/swiftlang/swift/issues/76242

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,153 @@
...
+    public func consumeWithInboundCloseHandler<Value: Sendable>(
+        _ operation: sending (RequestBody) async throws -> Value,
+        onInboundClosed: @Sendable @escaping () -> Void
+    ) async throws -> Value {
+        let iterator: UnsafeTransfer<NIOAsyncChannelInboundStream<HTTPRequestPart>.AsyncIterator> =
+            switch self._backing {
+            case .nioAsyncChannelRequestBody(let iterator):
+                iterator.underlyingIterator
+            default:
+                preconditionFailure("Cannot run consumeWithInboundCloseHandler on edited request body")
```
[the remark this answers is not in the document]
Joannis: I agree in-part, but crashing is a bit of an extreme measure here in my opinion. I'd be more open to an `assertionFailure`, but a precondition is harsh for this type of error.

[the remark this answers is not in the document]
Joannis: If we do want to keep the preconditionFailure, we really have to add a way for a user to know of a `RequestBody` is still a stream at all. I'd like not to expose that, but some middleware might collate your request body and you don't want to crash because of that.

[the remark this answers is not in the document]
adam-fowler: The other option is a keep a reference to the underlying iterator in the other cases so it can still be used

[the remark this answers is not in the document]
Joannis: That's a good option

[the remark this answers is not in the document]
adam-fowler: That doesn't work very well in the end. It's a bit clunky to have to keep a reference around and then you are iterating the wrong sequence for the request body ie the original one and not the one attached to the request now. I can probably add it for requests where the body has been collected into one buffer (as iteration is finished then), but if someone completely replaces the request body stream it just won't work.

sliemeobn:
something feels off here, I can quite put my finger down on it 🤔

I am a bit confused as to what the "signal" for client-side termination is supposed to be, because IIUC it is not exactly "end of request body", but more like the TCP-level "FIN". The whole thing is a bit of a protocol-mix-up mess, but I think that is the established norm (I could be totally off here), and I am not sure how that actually translates to HTTP/2 multiplexing.

But, I if the above is true, then collecting the request body to completion should have nothing to do with "the client closed the connection". And if the above is not true, can someone explain what is or share a link?

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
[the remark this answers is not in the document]
adam-fowler: I have other options for keeping the original async iterator around. But they are quite involved. So will look to do in a separate PR

adam-fowler:
> something feels off here, I can quite put my finger down on it 🤔
> 
> I am a bit confused as to what the "signal" for client-side termination is supposed to be, because IIUC it is not exactly "end of request body", but more like the TCP-level "FIN". The whole thing is a bit of a protocol-mix-up mess, but I think that is the established norm (I could be totally off here), and I am not sure how that actually translates to HTTP/2 multiplexing.
> 
> But, I if the above is true, then collecting the request body to completion should have nothing to do with "the client closed the connection". And if the above is not true, can someone explain what is or share a link?
> 
> edit: ok, I think I misunderstood the code, seems like it is doing just that...

The way it works is having a separate task that iterates the underlying request part stream. If this stream ends it indicates the connection has been closed and any more work should be cancelled. 

Lower down in the NIO code if an inbound closed message is received (ie the client sent a FIN) it finishes the inbound request part stream. This then triggers the calling of the inbound closed handler higher up in `withInboundCloseHandler`.  For HTTP2 the request part stream is closed when the client sends a `RST_STREAM`.

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
```diff
@@ -0,0 +1,144 @@
...
+    /// - Parameters
+    ///   - isolation: The isolation of the method. Defaults to the isolation of the caller.
+    ///   - operation: The actual operation to run
+    /// - Returns: Return value of operation
+    public func consumeWithCancellationOnInboundClose<Value: Sendable>(
+        _ operation: sending @escaping (RequestBody) async throws -> Value
+    ) async throws -> Value {
+        let (barrier, source) = AsyncStream<Void>.makeStream()
+        return try await consumeWithInboundCloseHandler { body in
+            try await withThrowingTaskGroup(of: Value.self) { group in
```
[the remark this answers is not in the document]
sliemeobn: Why though? sure, `~Escapable` would give us compile-time blessing, but if you don't pass the group anywhere, it will not escape. I don't see how these are connected, really.

sliemeobn:
thanks, that makes perfect sense. still will be great to "disconnect" the fate of collecting the body from losing the underlying close signal. but as you mentioned above, this can be done in a separate PR.

Tests/HummingbirdTests/ApplicationTests.swift
```diff
@@ -866,6 +866,34 @@ final class ApplicationTests: XCTestCase {
...
+    /// Test consumeWithInboundCloseHandler
+    func testConsumeWithInboundHandler() async throws {
+        let router = Router()
+        router.post("streaming") { request, context -> Response in
+            Response(
+                status: .ok,
+                body: .init { writer in
+                    try await request.body.consumeWithInboundCloseHandler { body in
+                        try await writer.write(body)
+                    } onInboundClosed: {
```
Joannis, line 880: What's being tested here?

Tests/HummingbirdCoreTests/CoreTests.swift
```diff
@@ -553,6 +553,125 @@ final class HummingBirdCoreTests: XCTestCase {
...
+                            try Task.checkCancellation()
+                            try await Task.sleep(for: .seconds(1))
+                            try await bodyWriter.write(body)
+                        } catch {
+                            throw error
+                        }
+                    }
+                    try await Task.sleep(for: .seconds(60))
+                }
+                try await bodyWriter.finish(nil)
```
Joannis, line 662: This shouldn't get triggered right? Can we `XCTFail()` here?

Joannis:
Question for longer-lived streams like SSE: Should we integrate those with graceful shutdown somehow?

sliemeobn:
just thinking out loud here:

what if we offered an initializer overload, something like
```swift
    router.get("/my-stream") { request, _ in
        return Response(
            status: .ok,
            headers: [.contentType: "text/event-stream"],
            body: .streamWithCancellation { writer in
                let mySource = someThing.subscribe()
                do {
                    for await event in mySource {
                        switch event {
                            // ...
                            writer.write(event)
                            // ...
                    }
                }
            }
        )
    }
```
with a better name somehow?

Tests/HummingbirdTests/ApplicationTests.swift
adam-fowler: That you can run consumeWithInboundCloseHandler and the request body passed in will still be correct.

Tests/HummingbirdCoreTests/CoreTests.swift
adam-fowler: done

adam-fowler:
> Question for longer-lived streams like SSE: Should we integrate those with graceful shutdown somehow?

So this is what I have in the SSE example 
```swift
    router.get("events") { request, context -> Response in
        .init(
            status: .ok,
            headers: [.contentType: "text/event-stream"],
            body: .init { writer in
                let allocator = ByteBufferAllocator()
                let (stream, id) = requestPublisher.subscribe()
                try await withGracefulShutdownHandler {
                    // If connection if closed then this function will call the `onInboundCLosed` closure
                    try await request.body.consumeWithInboundCloseHandler { request in
                        for try await value in stream {
                            try await writer.write(
                                ServerSentEvent(data: .init(string: value)).makeBuffer(
                                    allocator: allocator
                                )
                            )
                        }
                    } onInboundClosed: {
                        requestPublisher.unsubscribe(id)
                    }
                } onGracefulShutdown: {
                    requestPublisher.unsubscribe(id)
                }
                try await writer.finish(nil)
            }
        )
    }
```
@sliemeobn note I use the `consumeWithInboundCloseHandler` because that works just fine with the inout writer. It is a better solution as you aren't setting up addition child tasks and then sending a closure in an unsafe wrapper (which is required of `consumeWithCancellationOnInboundClose`). 

Your `.streamWithCancellation` will have exactly the same issues as `.consumeWithCancellationOnInboundClose` at some point I have to wrap a closure in an unsafe wrapper and convert a non-escaping closure to an escaping closure.

adam-fowler:
@Joannis wrapping inboundClosed and gracefulShutdown into one handler is perfectly doable. I think it should considered an addition to this PR, so dealt with separately. Is there anything left to resolve here? Regarding naming do you still prefer `read` over `consume`?

@sliemeobn `consumeWithCancellationOnInboundClose` I would like to investigate separately as well.

sliemeobn:
_>> It is a better solution as you aren't setting up addition child tasks and then sending a closure in an unsafe wrapper (which is required of consumeWithCancellationOnInboundClose)._
_>> consumeWithCancellationOnInboundClose I would like to investigate separately as well._

@adam-fowler I understand, but I think it is a bit unfortunate because currently you have to rely on having a "source" with an explicit `unsubscribe` or `cancel` function or similar at hand in your handler.
If you need to deal with an API that only vends an `AsyncSequence` and have to rely on task cancellation, you are in trouble again.

Also, I don't really see how using `withoutActuallyEscaping` would be "unsafe" here. IMHO the task group APIs are just "underbaked" for this use case, and `withoutActuallyEscaping` is the tool to get the "correct" API restored for the outside world. 

At least until somebody can explain to me what would be unsafe here, I shall remain stubborn on this point ^^

Sources/HummingbirdCore/Request/RequestBody+inboundClose.swift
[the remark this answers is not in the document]
Joannis: I agree with @sliemeobn here, we should consider `withoutActuallyEscaping`

Joannis (approved):
LGTM, we should handle @sliemeobn's comments in a second PR

PR #634, opened 2024-12-09
Storing underlying request part iterator with RequestBody

adam-fowler:
If you try to run `withInboundCloseHandler` on a request body that has been edited, there is a good chance you won't have a reference to the original request part iterator that is used to tell if the connection has been closed. If you did have a reference to the original iterator either it had iterated past the request body so couldn't be used to feed the request body in `withInboundCloseHandler` or it would provide the buffers from the original request and not the edited buffers.

To get around this we
- Store HTTPRequestPart iterator from original RequestBody with any subsequent request body setup.
- Using `RequestBody.mergeWithUnderlyingRequestPartIterator` we combine the current RequestBody with the original RequestPart iterator. With this we can supply the RequestBody buffer stream, but then also verify if the connection has been closed.

### DIFF
--- a/Tests/HummingbirdTests/ApplicationTests.swift
+++ b/Tests/HummingbirdTests/ApplicationTests.swift
@@ -893,6 +893,72 @@ final class ApplicationTests: XCTestCase {
             }
         }
     }
+
+    /// Test consumeWithInboundHandler
+    @available(macOS 15, iOS 18, tvOS 18, *)
+    func testConsumeWithInboundHandlerAfterCollect() async throws {
+        let router = Router()
+        router.post("streaming") { request, context -> Response in
+            var request = request
+            _ = try await request.collectBody(upTo: .max)
+            let request2 = request
+            return Response(
+                status: .ok,
+                body: .init { writer in
+                    try await request2.body.consumeWithInboundCloseHandler { body in
+                        try await writer.write(body)
+                    } onInboundClosed: {
+                    }
+                    try await writer.finish(nil)
+                }
+            )
+        }
+        let app = Application(responder: router.buildResponder())
+
+        try await app.test(.live) { client in
+            let buffer = Self.randomBuffer(size: 640_001)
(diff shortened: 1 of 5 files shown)


Tests/HummingbirdTests/ApplicationTests.swift
```diff
@@ -893,6 +893,72 @@ final class ApplicationTests: XCTestCase {
...
+                    return ByteBuffer(bytes: view)
+                }
+            )
+            let request2 = request
+            return Response(
+                status: .ok,
+                body: .init { writer in
+                    try await request2.body.consumeWithInboundCloseHandler { body in
+                        try await writer.write(body)
+                    } onInboundClosed: {
```
Joannis, line 945: Shouldn't we assert this gets hit (or not)?

Tests/HummingbirdTests/ApplicationTests.swift
```diff
@@ -893,6 +893,72 @@ final class ApplicationTests: XCTestCase {
...
+        router.post("streaming") { request, context -> Response in
+            var request = request
+            _ = try await request.collectBody(upTo: .max)
+            let request2 = request
+            return Response(
+                status: .ok,
+                body: .init { writer in
+                    try await request2.body.consumeWithInboundCloseHandler { body in
+                        try await writer.write(body)
+                    } onInboundClosed: {
```
Joannis, line 910: Same here, shouldn't we assert whether or not it gets hit?

Joannis (approved):

Tests/HummingbirdTests/ApplicationTests.swift
adam-fowler: No because the client never closes the connection

Tests/HummingbirdTests/ApplicationTests.swift
adam-fowler: No because the client never closes the connection

PR #637, opened 2024-12-20
Router validation

adam-fowler:
Add `Router.validate()` to check whether any route paths clash.

Unfortunately cannot add this in `buildResponder` as it is a throwing function.


Sources/Hummingbird/Router/Router+validation.swift
```diff
@@ -0,0 +1,111 @@
...
+            let count = min(lhs.path.count, rhs.path.count)
+            for i in 0..<count {
+                if lhs.path[i].priority < rhs.path[i].priority {
+                    return false
+                }
+            }
+            return true
+        }
+        guard trieValues.count > 1 else { return }
+        for index in 1..<trieValues.count {
```
Joannis, line 64: I think this is computationally much heavier than it needs to be, though it will catch more routes that overlap.
Let's break down what we want to validate (and why) in a discussion somewhere.

adam-fowler: I was intending to catch any route that would never be triggered because a higher priority route was triggered instead.

Sources/Hummingbird/Router/Router+validation.swift
```diff
@@ -0,0 +1,111 @@
...
+            endpoint.value.methods.keys
+                .sorted { $0.rawValue < $1.rawValue }
+                .map { RouteDescription(path: endpoint.path, method: $0) }
+        }
+    }
+
+    /// Validate router
+    ///
+    /// Verify that routes are not clashing
+    public func validate() throws {
```
Joannis, line 42: Would we recommend users to call this themselves? Or do we add an `#if DEBUG` block that calls this somewhere?

adam-fowler: It would have been good to add to the `Router.buildResponder` function but that isn't a throwing function. I was thinking of adding it to the template.

Joannis: What about adding this?

```swift
#if DEBUG
do {
  try self.validate()
} catch {
  assertionFailure("......")
}
#endif
```

adam-fowler: I guess I could

Sources/Hummingbird/Router/Router+validation.swift
adam-fowler: Optimised this. It now does the validation at the trie node level which is considerably faster

Joannis (approved):

PR #641, opened 2024-12-26
JSONDecode and NIOFoundationCompat

adam-fowler:
If we are to support just importing FoundationEssentials then we cannot include `NIOFoundationCompat` because it imports the full Foundation framework. This PR adds the elements of `NIOFoundationCompat` that we need to Hummingbird and it also makes the inclusion of NIOTransportServices conditional on building for Apple OS as that uses `NIOFoundationCompat`.

### DIFF
--- a/Sources/Hummingbird/Utils/ByteBuffer+foundation.swift
+++ b/Sources/Hummingbird/Utils/ByteBuffer+foundation.swift
@@ -36,7 +36,7 @@ import Foundation
 extension ByteBuffer {
     /// Controls how bytes are transferred between `ByteBuffer` and other storage types.
     @usableFromInline
-    package enum ByteTransferStrategy: Sendable {
+    package enum _ByteTransferStrategy: Sendable {
         /// Force a copy of the bytes.
         case copy
 
@@ -61,7 +61,7 @@ extension ByteBuffer {
     ///     - length: The number of bytes of interest
     /// - returns: A `Data` value containing the bytes of interest or `nil` if the selected bytes are not readable.
     @usableFromInline
-    package func getData(at index: Int, length: Int) -> Data? {
+    package func _getData(at index: Int, length: Int) -> Data? {
         self.getData(at: index, length: length, byteTransferStrategy: .automatic)
     }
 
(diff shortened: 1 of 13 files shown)


Sources/Hummingbird/Utils/ByteBuffer+foundation.swift
```diff
@@ -0,0 +1,172 @@
...
+    ///     - type: The type of the value to decode from the supplied JSON object.
+    ///     - buffer: The `ByteBuffer` that contains JSON object to decode.
+    /// - returns: The decoded object.
+    package func decodeByteBuffer<T: Decodable>(_ type: T.Type, from buffer: ByteBuffer) throws -> T {
+        try buffer._getJSONDecodable(
+            T.self,
+            decoder: self,
+            at: buffer.readerIndex,
+            length: buffer.readableBytes
+        )!  // must work, enough readable bytes// must work, enough readable bytes
```
Joannis, line 156:
```suggestion
        )!  // must work, enough readable bytes
```

Sources/Hummingbird/Utils/ByteBuffer+foundation.swift
```diff
@@ -0,0 +1,172 @@
...
+    package func _getJSONDecodable<T: Decodable>(
+        _ type: T.Type,
+        decoder: JSONDecoder = JSONDecoder(),
+        at index: Int,
+        length: Int
+    ) throws -> T? {
+        guard let data = self._getData(at: index, length: length, byteTransferStrategy: .noCopy) else {
+            return nil
+        }
+        return try decoder.decode(T.self, from: data)
```
Joannis, line 130: Can we do a `withData`-style API instead? The `noCopy` byteTransferStrategy seems like a bad idea to me, since it's easily miswielded.

Joannis (changes requested):

Sources/Hummingbird/Utils/ByteBuffer+foundation.swift
adam-fowler: These functions are direct copies of the functions in NIOFoundationCompat, except I added an `_` in front of them. The `ByteTransferStrategy.noCopy` works as far as I know. NIO seemed happy with it.

Joannis: That's fair, I don't think we need it here - but I agree we shouldn't diverse a "fork"

Joannis (approved):

PR #642, opened 2024-12-26
Follow symlinks when reading a file's attributes

### LINKED ISSUE #640: FileMiddleware provides an incorrect `Content-Length` when serving a file that is a symlink
To reproduce, on macOS:
- Place a file 1KB or more in size in a directory (e.g. `image.jpg`)
- Create a directory with a symlink to it (e.g., `ln -s image.jpg ./public/image.jpg`)
- Create a template project;
- Replace the default route with a `FileMiddleware` pointed to that directory (in the above example, `./public`);
- Try to fetch that file (e.g., `http://localhost…/image.jpg`).

The file will be truncated. The `Content-Size` returned matches the size for the symlink (as returned by `ls`), but not the size of the symlinked file. (The partial data will be from the symlinked file.)

Joannis:
Fixes #640

### DIFF
--- a/Sources/Hummingbird/Files/FileIO.swift
+++ b/Sources/Hummingbird/Files/FileIO.swift
@@ -42,7 +42,7 @@ public struct FileIO: Sendable {
         chunkLength: Int = NonBlockingFileIO.defaultChunkSize
     ) async throws -> ResponseBody {
         do {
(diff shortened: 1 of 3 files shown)


adam-fowler (commented):
Should we not also follow the symlink to get the file's contents

adam-fowler (approved):

PR #644, opened 2024-12-30
Fix active_requests metric name

adam-fowler:
Err somehow the metric name got set to "cod"?!?


Joannis (approved):
How did that happen?!

PR #649, opened 2025-01-07
Fix other http header dates (eg FileMiddleware output)

adam-fowler:


Sources/Hummingbird/Utils/DateCache+Date.swift
```diff
@@ -20,7 +20,7 @@ extension DateCache {
     static var rfc1123Formatter: DateFormatter {
         let formatter = DateFormatter()
         formatter.locale = Locale(identifier: "en_US_POSIX")
-        formatter.dateFormat = "EEE, d MMM yyy HH:mm:ss z"
+        formatter.dateFormat = "EEE, dd MMM yyy HH:mm:ss z"
```
Joannis, line 23: Do we need a test for this?

adam-fowler: I actually have another fix for this. Which is part of the FoundationEssentials work. I'll post that soon as an alternative

adam-fowler:
#650 replaces this

PR #652, opened 2025-01-13
Add URLEncodedFormDe/Encoder exception for URL to format as string

mredig:
Added `URL` alongside `Date` and `Data` as exceptions to the processing of URLEncodedFormDe/Encoders. Instead of treating a URL as a dictionary, it now matches the current behaviors of JSONDe/Encoder.

Tests were added and run before and after to confirm changes were successful.


adam-fowler (approved):
looks good. 
As long as CI is happy we are good to go

PR #655, opened 2025-01-13
Use NIOFileSystem

adam-fowler:
I've set this as a draft for the moment as while NonBlockingFileIO has been deprecated, NIOFileSystem still has an under bar prefix which indicates it has an unstable API.


Joannis (approved):

.github/workflows/ci.yml
Joannis, line 43: Is this no longer needed?

Joannis (approved):

.github/workflows/ci.yml
adam-fowler: It will get re-enabled when swift 6.4 works with benchmark package

PR #657, opened 2025-01-15
Add ability for developers to support additional MIME types in FileMiddleware

### LINKED ISSUE #656: MediaType isn't extensible
I am currently working on an HLS server and was working to serve the stream components through HB, however when I load an `m3u8` file through `FileMiddleware`, there's no definition for `m3u8` in `MediaType` so the browser just attempts to load it as plain text.

Obviously, the simplest solution would be to just add `m3u8` to `MediaType.extensionMap`, but I think that's just a band aid. Devs should be able to define additional MediaTypes that can be used by FileMiddleware.

I would propose that `MediaType.extensionMap` be made privately mutable and to add a static `registerAdditionalType(_ type: MediaType, withExtension ext: String)` to `MediaType` (or, if scoping is preferable for some reason I can't foresee, perhaps registering on an instance of `FileMiddleware`)

I'd be happy to make a PR for this if you guys are onboard.

mredig:
Added `additionalMediaTypeExtensions` to `FileMiddleware` with a `withAdditionalMediaType` method, adding additional media types, to retain object immutability.

I also went out on a limb and funneled the `FileMiddleware` inits through a single initializer as that better supports this pattern.

This would resolve #656

A test case was added to confirm expected behavior.

### DIFF
--- a/Sources/Hummingbird/Middleware/FileMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/FileMiddleware.swift
@@ -52,6 +52,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
     let searchForIndexHtml: Bool
     let urlBasePath: String?
     let fileProvider: Provider
+    let additionalMediaTypeExtensions: [String: MediaType]
 
     /// Create FileMiddleware
     /// - Parameters:
@@ -69,13 +70,16 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
         threadPool: NIOThreadPool = NIOThreadPool.singleton,
         logger: Logger = Logger(label: "FileMiddleware")
     ) where Provider == LocalFileSystem {
-        self.cacheControl = cacheControl
-        self.searchForIndexHtml = searchForIndexHtml
-        self.urlBasePath = urlBasePath.map { String($0.dropSuffix("/")) }
-        self.fileProvider = LocalFileSystem(
-            rootFolder: rootFolder,
-            threadPool: threadPool,
-            logger: logger
+        self.init(
+            fileProvider: LocalFileSystem(
+                rootFolder: rootFolder,
+                threadPool: threadPool,
+                logger: logger
+            ),
+            urlBasePath: urlBasePath,
+            cacheControl: cacheControl,
+            searchForIndexHtml: searchForIndexHtml,
+            additionalMediaTypeExtensions: [:]
         )
     }
 
@@ -91,10 +95,39 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
         cacheControl: CacheControl = .init([]),
         searchForIndexHtml: Bool = false
     ) {
+        self.init(
+            fileProvider: fileProvider,
+            urlBasePath: urlBasePath,
+            cacheControl: cacheControl,
+            searchForIndexHtml: searchForIndexHtml,
+            additionalMediaTypeExtensions: [:]
+        )
+    }
+
+    private init(
+        fileProvider: Provider,
+        urlBasePath: String? = nil,
+        cacheControl: CacheControl = .init([]),
+        searchForIndexHtml: Bool = false,
+        additionalMediaTypeExtensions: [String: MediaType]
+    ) {
         self.cacheControl = cacheControl
         self.searchForIndexHtml = searchForIndexHtml
         self.urlBasePath = urlBasePath.map { String($0.dropSuffix("/")) }
         self.fileProvider = fileProvider
+        self.additionalMediaTypeExtensions = additionalMediaTypeExtensions
+    }
+
+    public func withAdditionalMediaType(_ mediaType: MediaType, forFileExtension fileExtension: String) -> FileMiddleware {
+        var extensions = additionalMediaTypeExtensions
+        extensions[fileExtension] = mediaType
+        return FileMiddleware(
+            fileProvider: fileProvider,
+            urlBasePath: urlBasePath,
+            cacheControl: cacheControl,
+            searchForIndexHtml: searchForIndexHtml,
+            additionalMediaTypeExtensions: extensions
+        )
     }
 
     /// Handle request
@@ -220,7 +253,7 @@ extension FileMiddleware {
 
         // content-type
         if let ext = self.fileExtension(for: path) {
-            if let contentType = MediaType.getMediaType(forExtension: ext) {
+            if let contentType = additionalMediaTypeExtensions[ext] ?? MediaType.getMediaType(forExtension: ext) {
                 headers[.contentType] = contentType.description
             }
         }
--- a/Sources/Hummingbird/Middleware/FileMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/FileMiddleware.swift
@@ -52,7 +52,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
     let searchForIndexHtml: Bool
     let urlBasePath: String?
     let fileProvider: Provider
-    let additionalMediaTypeExtensions: [String: MediaType]
+    let mediaTypeFileExtensionMap: [String: MediaType]
 
     /// Create FileMiddleware
     /// - Parameters:
@@ -79,7 +79,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
             urlBasePath: urlBasePath,
             cacheControl: cacheControl,
             searchForIndexHtml: searchForIndexHtml,
-            additionalMediaTypeExtensions: [:]
+            mediaTypeFileExtensionMap: [:]
         )
     }
 
@@ -100,7 +100,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
             urlBasePath: urlBasePath,
             cacheControl: cacheControl,
             searchForIndexHtml: searchForIndexHtml,
-            additionalMediaTypeExtensions: [:]
+            mediaTypeFileExtensionMap: [:]
         )
     }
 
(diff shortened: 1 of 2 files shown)


Joannis:
I actually think it's a good case for making that property explicitly mutable. I don't see any reason we wouldn't do so. While adding a function to modify those properties is nice, it's certainly also no harm and potentially helpful to know all registered file extensions.

Joannis:
That makes sense!

mredig:
> That makes sense!

So, leave it as it is, but make the property public?

adam-fowler:
> I actually think it's a good case for making that property explicitly mutable. I don't see any reason we wouldn't do so. While adding a function to modify those properties is nice, it's certainly also no harm and potentially helpful to know all registered file extensions.

I don't think it's a great idea to make what should be a static list (you aren't going to change this while the server is running) mutable.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -69,13 +70,16 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
+        self.init(
+            fileProvider: LocalFileSystem(
+                rootFolder: rootFolder,
+                threadPool: threadPool,
+                logger: logger
+            ),
+            urlBasePath: urlBasePath,
+            cacheControl: cacheControl,
+            searchForIndexHtml: searchForIndexHtml,
+            additionalMediaTypeExtensions: [:]
```
adam-fowler, line 82: mediaTypeFileExtensionMap?

mredig: will do!

adam-fowler: Of should it be the other way fileExtensionMediaTypeMap? What do you think

mredig: I think they are about equal in terms of clarity, but the first is more concise.

mredig: > I think they are about equal in terms of clarity, but the first is more concise.

I just realized they are the same character count, but the second *looks* longer lol

mredig: Just pushed - I also changed 
`public func withAdditionalMediaType(_ mediaType: MediaType, forFileExtension fileExtension: String) -> FileMiddleware`
to 
`public func withAdditionalMediaType(_ mediaType: MediaType, mappedToFileExtension fileExtension: String) -> FileMiddleware`
to match.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -52,6 +52,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
     let searchForIndexHtml: Bool
     let urlBasePath: String?
     let fileProvider: Provider
+    public let mediaTypeFileExtensionMap: [String: MediaType]
```
adam-fowler, line 55: Is there a reason this needs to be public.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -90,11 +94,40 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
+        mediaTypeFileExtensionMap: [String: MediaType]
     ) {
         self.cacheControl = cacheControl
         self.searchForIndexHtml = searchForIndexHtml
         self.urlBasePath = urlBasePath.map { String($0.dropSuffix("/")) }
         self.fileProvider = fileProvider
+        self.mediaTypeFileExtensionMap = mediaTypeFileExtensionMap
+    }
+
+    public func withAdditionalMediaType(_ mediaType: MediaType, mappedToFileExtension fileExtension: String) -> FileMiddleware {
```
adam-fowler, line 121: Is it worthwhile adding a function to add multiple media types. 
```
public func withAdditionalMediaTypes(_ types: [String: MediaType]) -> FileMiddleware
```

adam-fowler (commented):
Couple of comments but otherwise looks good

Sources/Hummingbird/Middleware/FileMiddleware.swift
mredig: re @Joannis 
>it's certainly also no harm and potentially helpful to know all registered file extensions.

I thought it was a good point. I could see a need to introspect what the customizations are.

Sources/Hummingbird/Middleware/FileMiddleware.swift
mredig: Here are sample call sites:

```swift
FileMiddleware(".")
    .withAdditionalMediaType(hlsStream, mappedToFileExtension: "m3u8")
    .withAdditionalMediaType(MediaType(type: .application, subType: "foo"), mappedToFileExtension: "foo")
    .withAdditionalMediaType(MediaType(type: .application, subType: "bar"), mappedToFileExtension: "bar")
    .withAdditionalMediaType(MediaType(type: .application, subType: "baz"), mappedToFileExtension: "baz")
```

```swift
FileMiddleware(".")
    .withAdditionalMediaTypes([
        "m3u8": hlsStream,
        "foo": MediaType(type: .application, subType: "foo"),
        "bar": MediaType(type: .application, subType: "bar"),
        "baz": MediaType(type: .application, subType: "baz"),
    ])
```

I definitely like the less imposing block of text, but it still retains the original issue I was trying to avoid in that it's not super apparent that the keys in the dictionary are file extensions.

>...future maintainers would have the explicit context of associating the file extension with the mime type instead of having to intuit the association.

What would you think of

`public func withAdditionalMediaTypes(forFileExtensions extensionToMediaTypeMap: [String: MediaType]) -> FileMiddleware`

Sources/Hummingbird/Middleware/FileMiddleware.swift
adam-fowler: Until someone needs it I'd prefer we leave it private. Setting it to public means setting it in stone until the next major version.

Sources/Hummingbird/Middleware/FileMiddleware.swift
adam-fowler: Sounds good

mredig: I'll get on it!

mredig:
I looked at the details for the validation failure and it isn't making sense to me. Does it want me to format like this?!?!

```swift
let extensions = 
    extensionToMediaTypeMap
    .reduce(into: mediaTypeFileExtensionMap) {
        $0[$1.key.lowercased()] = $1.value
    }
```

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -90,11 +94,47 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
+        withAdditionalMediaTypes(forFileExtensions: [fileExtension: mediaType])
+    }
+
+    public func withAdditionalMediaTypes(forFileExtensions extensionToMediaTypeMap: [String: MediaType]) -> FileMiddleware {
```
Joannis, line 125: @adam-fowler Is it worth adding a custom `ExpressibleByDictionaryLiteral` type for this dictionary use case? That way we can represent it in, what I believe, would be a clearer API surface. That could also cascade into the `public var extensionMap: ... { get }` down the line

adam-fowler: We can look at this in another PR. We can merge this, but if we want a custom type we'll need that PR before we do a release.

adam-fowler (approved):
I'm happy with this as is. But need to run swift-format on the files.

Sources/Hummingbird/Middleware/FileMiddleware.swift
Joannis: @mredig do you want to give that custom type a shot? I think it'd be a great addition.

mredig: Yeah I can do another PR as @adam-fowler suggested, after this one. :)

mredig:
> I'm happy with this as is. But need to run swift-format on the files.

I just ran it and pushed the change, but I'm skeptical that this change is desirable.

```
     public func withAdditionalMediaTypes(forFileExtensions extensionToMediaTypeMap: [String: MediaType]) -> FileMiddleware {
-        let extensions = extensionToMediaTypeMap
+        let extensions =
+            extensionToMediaTypeMap
             .reduce(into: mediaTypeFileExtensionMap) {
                 $0[$1.key.lowercased()] = $1.value
             }
```

mredig:
> > I'm happy with this as is. But need to run swift-format on the files.
> 
> I just ran it and pushed the change, but I'm skeptical that this change is desirable.
> 
> ```
>      public func withAdditionalMediaTypes(forFileExtensions extensionToMediaTypeMap: [String: MediaType]) -> FileMiddleware {
> -        let extensions = extensionToMediaTypeMap
> +        let extensions =
> +            extensionToMediaTypeMap
>              .reduce(into: mediaTypeFileExtensionMap) {
>                  $0[$1.key.lowercased()] = $1.value
>              }
> ```

I dug into swift-format yesterday and found that this appears to be intentional formatting. I suppose it is what it is, but ...





(Not trying to change your mind or anything - I just have strong opinions on formatting)

adam-fowler:
I'm not necessarily happy about some of the decisions swift-format makes around breaking up lines. 

But I decided to go with swift-format to ease access to the project for all developers. Swift-format is packaged with the compiler.

You can sometimes get around some the decisions by breaking up lines as you'd prefer. In this case I'm sure you could merge the `reduce` line back into the previous line and get something a little more attractive.

adam-fowler:
I change the formatting to be slightly more desirable and got this merged

mredig:
> I change the formatting to be slightly more desirable and got this merged

lolol thank you. Honestly, I was just complaining, but expected it to just be used as-was. Looks better though! :D

PR #658, opened 2025-01-15
Show URLFormCoding test case errors in test case

mredig:
While making the URL en/decoding PR, I noticed that test failures showed inline alerts at the top of the code file, not in the test case itself, due to the convenience methods designed for the tests.

I adjusted the definitions of these convenience methods to instead pass the file and line of the actual test and forward that to `XCTAssert*`, showing errors inline at the call site instead of stacking failures in a single location.

Incidentally, I also noticed that all the tests are still using the XCTest framework. Is there interest in migrating to Swift Testing? That would be a much more time consuming undertaking, so I don't want to make any PRs on that without any demonstrated interest.

### DIFF
--- a/Tests/HummingbirdTests/URLEncodedForm/URLDecoderTests.swift
+++ b/Tests/HummingbirdTests/URLEncodedForm/URLDecoderTests.swift
@@ -17,12 +17,18 @@ import XCTest
 @testable import Hummingbird
 
 final class URLDecodedFormDecoderTests: XCTestCase {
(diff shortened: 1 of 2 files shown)


Joannis:
IIRC, we didn't move to Swift Testing because of our support for older Swift releases.

Joannis (approved):

adam-fowler (approved):

mredig:
> IIRC, we didn't move to Swift Testing because of our support for older Swift releases.

Ah. That makes sense. I suppose you'd need to test against all supported versions of Swift and not just the latest.

PR #659, opened 2025-01-16
Fix iOS build: NIOTSEventLoopGroup's module is not imported

millenomi:
Building for iOS fails with a single error in this file.

Fix by explicitly using `import NIOTransportServices` on use.

### DIFF
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -17,6 +17,7 @@ import Logging
 import NIOCore
 import NIOHTTPTypes
 import NIOPosix
+import NIOTransportServices
 import ServiceLifecycle
 import UnixSignals
 
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -17,9 +17,14 @@ import Logging
 import NIOCore
 import NIOHTTPTypes
 import NIOPosix
-import NIOTransportServices
 import ServiceLifecycle
 import UnixSignals
+
+#if os(iOS) && canImport(NIOTransportServices)
+// Import for NIOTSEventLoopGroup, which we're accessing
+// with #if os(iOS), below.
+import NIOTransportServices
+#endif
 
 /// Where should the application get its EventLoopGroup from
 public enum EventLoopGroupProvider {
--- a/Sources/Hummingbird/Application.swift
+++ b/Sources/Hummingbird/Application.swift
@@ -20,9 +20,7 @@ import NIOPosix
 import ServiceLifecycle
 import UnixSignals
 
-#if os(iOS) && canImport(NIOTransportServices)
-// Import for NIOTSEventLoopGroup, which we're accessing
-// with #if os(iOS), below.
+#if os(iOS)
 import NIOTransportServices
 #endif
 
@@ -36,7 +34,7 @@ public enum EventLoopGroupProvider {
     public var eventLoopGroup: EventLoopGroup {
         switch self {
         case .singleton:
-            #if os(iOS) && canImport(NIOTransportServices)
+            #if os(iOS)
             return NIOTSEventLoopGroup.singleton
             #else
             return MultiThreadedEventLoopGroup.singleton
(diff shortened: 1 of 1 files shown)


Joannis:
Hey @millenomi, I think it's a good idea to add an iOS CI.

adam-fowler (changes requested):
That'll break non-apple OSs as NIOTransportServices is not dependency for them, although CI doesn't seem to mind

millenomi:
Good point. I'll use canImport separately.

millenomi:
I have guarded the import both with `canImport` and `os(iOS)`, to match the symbol usage and not introduce dependency changes on other platforms.

A note here may be that `os(iOS)` does _not_ cover the entire family of iOS-derived platforms. i.e., building this will go in the `#else` path on tvOS, visionOS and watchOS. One way I handle this kind of mess without continual repetition is usually by defining a flag in Package.swift, which here would look something like:

```swift
let nioTransportServicesPlatforms = [
    Platform.iOS,
        .tvOS,
        .watchOS,
        .visionOS,
]

…

dependencies: [
    .package(…, .when(platforms: nioTransportServicesPlatforms))
],
…
// In the target that uses it:
swiftSettings: [
    .define("HUMMINGBIRD_USES_NIOTRANSPORTSERVICES", .when(platforms: nioTransportServicesPlatforms))
]
```

Then the use would become:

```swift
#if HUMMINGBIRD_USES_NIOTRANSPORTSERVICES
import NIOTransportServices
#endif
```

But I see that there are other issues building (e.g. for watchOS) right now, so I opted to keep the patch as close as possible to the original while also fixing the build issue.

Sources/Hummingbird/Application.swift
```diff
@@ -20,6 +20,12 @@ import NIOPosix
 import ServiceLifecycle
 import UnixSignals
 
+#if os(iOS) && canImport(NIOTransportServices)
```
adam-fowler, line 23: I've always used `canImport(Network)`. You don't need the platform check. 

CanImport on SwiftPM modules is decidedly flakey.

adam-fowler: Actually looking at the code it is only iOS that uses `NIOTSEventLoopGroup` by default.
This would do
```
#if os(iOS)
import NIOTransportServices
#endif
```

You'll need to run swift-format on the file after as swift-format orders all conditional imports after non-conditional ones

adam-fowler:
> A note here may be that `os(iOS)` does _not_ cover the entire family

But it is the only platform that references NIOTransportService symbols in that file.

watchOS is not supported. Apple only allow network access to watches in very limited situations

millenomi:
> But it is the only platform that references NIOTransportService symbols in that file.

Yep. Other iOS derivative platforms will be using the other singleton, which according to NIO's docs is not recommended for use.

I'm deploying specifically on iOS at the moment, so it's not a priority _for me_ specifically, but if the project attempts to support iOS-family OSes it's something that perhaps ought to be fixed. Would you like an issue to be filed wrt that?

> watchOS is not supported. Apple only allow network access to watches in very limited situations

Oh, whoops, this one's my bad, Please disregard my watchOS tangent — I thought I had seen watchOS in the manifest somewhere and it is in fact not mentioned anywhere, correctly. I blame my inattention and Xcode defaulting to my watch as the deployment destination for newly opened projects, for some reason.

Sources/Hummingbird/Application.swift
millenomi: On it, thank you.

millenomi: Adopted in [Simplify back to just #if os(iOS). Checked with a swift-format run …](https://github.com/hummingbird-project/hummingbird/pull/659/commits/c1d328c87597beb2fb16a264f0d8614940b2f432).

millenomi:
Looks like there's an infra issue with the benchmark test?

adam-fowler (approved):

adam-fowler:
> Looks like there's an infra issue with the benchmark test? [Details](https://github.com/hummingbird-project/hummingbird/actions/runs/12835649407/job/35798257091?pr=659)

Yeah that's an issue with PRs posted by people outside of the hummingbird org. Haven't worked out how to resolve it yet.

PR #661, opened 2025-01-18
FileMiddleware will provide correct mime type for non lowercase file extensions

mredig:
I added a test for `FileMiddleware` to test handling file extensions that aren't homogeneous lowercase and it failed.

I then updated `FileMiddlware` to normalize as lowercase the file extensions it extracts from filenames (to match the existing convention on `MediaType`) prior to passing them to `MediaType.getMediaType(forExtension:)`.

The test passes now.


Joannis (approved):

Tests/HummingbirdTests/FileMiddlewareTests.swift
```diff
@@ -512,4 +512,33 @@ final class FileMiddlewareTests: XCTestCase {
...
+        let testedExtensions: [(String, UInt)] = [
+            ("jpg", #line),
+            ("JPG", #line),
+            ("JpG", #line),
+            ("JPeG", #line),
+            ("JPEG", #line),
+        ]
+
+        for (index, (testedExtension, line)) in testedExtensions.enumerated() {
+            let fileURL = URL(filePath: "\(#function)-\(index)")
```
adam-fowler, line 530: This isn't compiling under Linux 5.x. You can use `fileURLWithPath` instead. At least that is what I'm using above.

mredig: Is there a docker image to easily test this against?

mredig: > This isn't compiling under Linux 5.x. 

I just looked at the failed actions and realized that you meant Swift 5.x on linux... I wasn't sure how to get an image with linux kernel 5.x, especially given that containers share the kernel with the host 😅

mredig:
> Is it worth adding a custom ExpressibleByDictionaryLiteral type for this dictionary use case? That way we can represent it in, what I believe, would be a clearer API surface. [ref](https://github.com/hummingbird-project/hummingbird/pull/657#discussion_r1921031334)

I have the beginnings of this specific idea right here:
```swift
extension MediaType {
    public struct WellknownExtensions: RawRepresentable, Sendable, ExpressibleByDictionaryLiteral {
        public private(set) var rawValue: [FileExtension: MediaType]

        public init(rawValue: [MediaType.FileExtension : MediaType]) {
            self.rawValue = rawValue
        }

        public init(dictionaryLiteral elements: (FileExtension, MediaType)...) {
            self.init(rawValue: elements.reduce(into: .init(), {
                $0[$1.0] = $1.1
            }))
        }
    }
}
```

but it occurred to me that between the changes I already made by adding `FileExtension` and making it case-insensitive equatable, `ExpressibleByStringLiteral` and then naturally being able to use the existing interface with a dictionary literal argument, I can't think of any value this would bring.

I personally think that using a type safe wrapper for file extensions is also a cleaner interface for case insensitive comparison rather than something like `WellknownExtensions.compare("jpg", "JPG")`

Should I continue forward with this, or do you agree what I already have is better?

mredig:
Re - the api breaking changes...

I wasn't aware that the last PR had been released yet, but I can make sure there are `String` args complementing the `FileExtension` ones, if we want to proceed with this.

adam-fowler:
@mredig ah that was my bad.

Tests/HummingbirdTests/FileMiddlewareTests.swift
adam-fowler: You can use any `swift:5.x` docker image to test. I tend to use a devcontainer in VSCode to test these kinda things.

mredig: Gotcha. That's what I ended up doing too :)

As for devcontainers, I also heard you mention that in a podcast. I assume that's a VSCode extension or whatever they call it? I'm using Xcode for all this so far and, while I'm happy with it, I'm not against seeing what alternatives can offer.

mredig:
I am considering this currently as unapproved since I made significant changes since the approval, however unless you guys see any shortcomings, I am considering it complete and ready for real PR approval.

I just want to make sure that there's not a limbo of me waiting for you while you are waiting for something from me :)

mredig:
(the test failures are from certificates that expired hours prior to the tests running - in the `Tests/HummingbirdHTTP2Tests/Certificates.swift` file.)

.mailmap
```diff
@@ -2,4 +2,5 @@ Nicholas Trienens <nick@fuzzproductions.com> Nicholas Trienens <nick.trienens@mo
 Adam Fowler <adamfowler71@gmail.com> adam-fowler <adam-fowler@users.noreply.github.com>
 Adam Fowler <adamfowler71@gmail.com> dependabot[bot] <49699333+dependabot[bot]@users.noreply.github.com>
 Adam Fowler <adamfowler71@gmail.com> hummingbird-automation[bot] <168462326+hummingbird-automation[bot]@users.noreply.github.com>
-Michael <mredig@gmail.com> Michael <juniper.fife_0b@icloud.com>
\ No newline at end of file
+Michael Redig <juniper.fife_0b@icloud.com>  Michael <mredig@gmail.com>
```
adam-fowler, line 5: Is this your preferred email then?
Run ./scripts/generate-contributors.sh to update the contributors.md

Sources/Hummingbird/HTTP/MediaType+FileExtension.swift
```diff
@@ -0,0 +1,97 @@
+import HummingbirdCore
+
+extension MediaType {
+    public struct FileExtension: RawRepresentable, ExpressibleByStringLiteral, Sendable, Hashable {
```
adam-fowler, line 4: Public types need documentation as do their public members, methods

Tests/HummingbirdTests/FileMiddlewareTests.swift
```diff
@@ -513,6 +513,35 @@ final class FileMiddlewareTests: XCTestCase {
...
+
+        for (index, (testedExtension, line)) in testedExtensions.enumerated() {
+            let fileURL = URL(fileURLWithPath: "\(#function)-\(index)")
+                .appendingPathExtension(testedExtension)
+            let filename = fileURL.lastPathComponent
+            let data = Data()
+            XCTAssertNoThrow(try data.write(to: fileURL))
+            defer { XCTAssertNoThrow(try FileManager.default.removeItem(at: fileURL)) }
+
+            try await app.test(.router) { client in
```
adam-fowler, line 537: Instead of starting the app up five times, why don't you stick the for loop inside the `app.test()` call

adam-fowler (changes requested):
Couple of minor things, but in general looks good. I'm not sure it is necessary to include all the extensions as static members of `MediaType.FileExtension` but I'm not opposed either.

Note #665 includes updated certificates, so tests should work once that is merged

.mailmap
mredig: It's my hide-my-email address for GitHub. I know it's not doing much given the publicized nature of my other email in commits, etc, but it seems silly to not at least try to obscure it a little bit.

Will do!

Sources/Hummingbird/HTTP/MediaType+FileExtension.swift
mredig: I'll leave resolving conversations to you to make sure the changes are to your specifications.

adam-fowler (approved):

adam-fowler:
thanks Michael

PR #666, opened 2025-01-27
Combine RequestBody delegate, source state to fix hang

adam-fowler:
The waitingForProduceMore continuations and the produceMore state have been combined into one state in the Delegate of RequestBody, 

Fixes issue with hang where yield was told to stop producing and but then was told it could continue producing before the next call to yield.

An alternative solution, which is simpler, would be to wait on the continuation as soon as `yield` returns `stopProducing` but this means `Source.yield()` stalls on the yield that returns stop producing and not the next yield.

### DIFF
--- a/Sources/HummingbirdCore/Request/RequestBody.swift
+++ b/Sources/HummingbirdCore/Request/RequestBody.swift
@@ -134,27 +134,53 @@ extension RequestBody {
     /// Delegate for NIOThrowingAsyncSequenceProducer
     @usableFromInline
     final class Delegate: NIOAsyncSequenceProducerDelegate, Sendable {
-        let checkedContinuations: NIOLockedValueBox<Deque<CheckedContinuation<Void, Never>>>
+        enum State {
+            case produceMore
+            case waitingForProduceMore(Deque<CheckedContinuation<Void, Never>>)
+            case terminated
+        }
+        let state: NIOLockedValueBox<State>
 
         @usableFromInline
         init() {
-            self.checkedContinuations = .init([])
+            self.state = .init(.produceMore)
         }
 
         @usableFromInline
         func produceMore() {
-            self.checkedContinuations.withLockedValue {
-                if let cont = $0.popFirst() {
-                    cont.resume()
+            
+            self.state.withLockedValue { state in
+                switch state {
+                case .produceMore:
+                    break
+                case .waitingForProduceMore(var continuations):
+                    if let cont = continuations.popFirst() {
+                        cont.resume()
+                    }
+                    if continuations.count > 0 {
+                        state = .waitingForProduceMore(continuations)
+                    } else {
+                        state = .produceMore
+                    }
+                case .terminated:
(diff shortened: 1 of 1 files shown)


Sources/HummingbirdCore/Request/RequestBody.swift
```diff
@@ -134,36 +134,84 @@ extension RequestBody {
...
+                    break
+                case .waitingForProduceMore(var continuations):
+                    if let cont = continuations.popFirst() {
+                        cont.resume()
+                    }
+                    if continuations.count > 0 {
+                        state = .waitingForProduceMore(continuations)
+                    } else {
+                        state = .produceMore
+                    }
```
Joannis, line 164: I expect this code is quite performance sensitive, can we restructure to save a copy?

```suggestion
                    switch continuations.count {
                    case 0:
                        assertionFailure("`waitingToProduceMore` state with a continuation count of zero")
                        state = .produceMore
                    case 1:
                        continuations[0].resume()
                        state = .produceMore
                    default:
                        continuations.removeFirst().resume()
                        state = .waitingForProduceMore(continuations)
                    }
```

Joannis (approved):
LGTM, one performance remark here.

Sources/HummingbirdCore/Request/RequestBody.swift
adam-fowler: What about having two states, one for a single continuation (The most likely case) and one for multiple continuations.

Sources/HummingbirdCore/Request/RequestBody.swift
```diff
@@ -134,36 +134,88 @@ extension RequestBody {
...
             }
         }
 
         @usableFromInline
         func waitForProduceMore() async {
-            await withCheckedContinuation { (cont: CheckedContinuation<Void, Never>) in
-                self.checkedContinuations.withLockedValue {
-                    $0.append(cont)
+            switch self.state.withLockedValue({ $0 }) {
+            case .produceMore, .terminated:
```
adam-fowler, line 190: @Joannis this is an attempt to optimise waitForProduceMore to not create a continuation if the state is produceMore (the most probably state)

Sources/HummingbirdCore/Request/RequestBody.swift
adam-fowler: Would mean you wouldn't have the extra allocation for the deque contents

Sources/HummingbirdCore/Request/RequestBody.swift
Joannis: That makes sense! Great catch

Sources/HummingbirdCore/Request/RequestBody.swift
Joannis: That makes sense, my goal was indeed to omit the Deque allocation

Joannis (approved):

PR #673, opened 2025-02-08
Add custom verification handler to NIOSSLServerHandler

adam-fowler:


Sources/HummingbirdHTTP2/TLSChannelConfiguration.swift
```diff
@@ -0,0 +1,57 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+//===----------------------------------------------------------------------===//
+
+import NIOCore
+import NIOSSL
+
+/// TLSChannel configuration
+public struct TLSChannelConfiguration: Sendable {
+    public typealias CustomVerificationCallback = @Sendable ([NIOSSLCertificate], EventLoopPromise<NIOSSLVerificationResult>) -> Void
```
Joannis, line 20: Should we expose an API here that has a signature of `@Sendable ([NIOSSLCertificate]) async throws -> NIOSSLVerificationResult`?

adam-fowler: Had a look at this and that isn't what is expected of this function. If you check the function it sends in a promise, kicks off the custom verification process and returns immediately. Changing it to async/await means it will wait until the custom verification has completed. I could kick off an unstructured task inside the function. Not sure that's ideal though.

adam-fowler: In actual fact given how promise.completeWithTask works it doesnt really make much difference if it is an unstructured task. I have added an async verification callback, but left the EL version in as well

Joannis (approved):
Awesome! This looks great.

PR #674, opened 2025-02-10
Add Environment.require(_:), require(_:as:)

adam-fowler:
In most cases you are getting an environment variable you actually require it to exist.


Joannis (approved):
Surprised we didn't have this yet! I thought we did.

PR #679, opened 2025-02-24
Use TaskLocals to isolate metrics/tracing tests from each other

adam-fowler:
By isolating a tracer or metrics factory for each test we can run them in parallel. This isn't an issue at the moment really, but when we move to Swift Testing it will be so lets just fix it now.


ktoso (approved):
Nicely done :)

Joannis (approved):

PR #680, opened 2025-02-27
Add CacheControl tests

Cyberbeni:
Noticed that CacheControl doesn't work with .any, tried adding XCTExpectedFailure for the failing test but apparently it is not implemented on Linux. Don't know when I'll have the time to look into fixing it, so I'm only submitting the tests for now.

### DIFF
--- a/Sources/Hummingbird/HTTP/MediaType.swift
+++ b/Sources/Hummingbird/HTTP/MediaType.swift
@@ -168,8 +168,7 @@ public struct MediaType: Sendable, CustomStringConvertible {
 
         static func ~= (_ lhs: Self, _ rhs: Self) -> Bool {
             switch (lhs, rhs) {
-            case (.any, _),
-                (_, .any):
+            case (.any, _), (_, .any):
                 true
             default:
                 lhs == rhs
--- a/Sources/Hummingbird/HTTP/MediaType.swift
+++ b/Sources/Hummingbird/HTTP/MediaType.swift
@@ -154,7 +154,7 @@ public struct MediaType: Sendable, CustomStringConvertible {
 
     /// Media type categories
     public struct Category: Sendable, Equatable, RawRepresentable, CustomStringConvertible {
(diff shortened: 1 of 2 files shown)


adam-fowler:
I added a fix so you PR passes all its tests

Cyberbeni:
Removed the `func ==` implementation, since it wasn't working (probably because Category is RawRepresentable and that default implementation of Equatable is higher priority than default implementation for types with only equatable variables) and it also wasn't transitive. https://developer.apple.com/documentation/swift/equatable

> Since equality between instances of Equatable types is an equivalence relation, any of your custom types that conform to Equatable must satisfy three conditions, for any values a, b, and c:
>
> - a == a is always true (Reflexivity)
> - a == b implies b == a (Symmetry)
> - a == b and b == c implies a == c (Transitivity)

adam-fowler:
An alternative solution would be to change the `==` implementation to `~=` and remove the Equatable conformance from Category and use the `~=` inside `isType`.

Cyberbeni:
Ok, I also removed Category.Internal as it was redundant to have the same cases for the wrapper type.

Cyberbeni:
As far as I know, the only breaking change is that a switch case over Category will produce a warning for the default case if all the cases are listed.

adam-fowler:
> Ok, I also removed Category.Internal as it was redundant to have the same cases for the wrapper type.

Public enums in libraries are not a good idea, because adding cases to them is a breaking change. You need to wrap them in a struct to avoid the breaking change. See this pitch which will hopefully make it into swift https://forums.swift.org/t/pitch-extensible-enums-for-non-resilient-modules/77649

Cyberbeni:
Ah,this comes up every time I forget about it... It was even mentioned in the original proposal: https://github.com/swiftlang/swift-evolution/blob/main/proposals/0192-non-exhaustive-enums.md#allow-enums-defined-in-source-packages-to-be-considered-non-frozen

adam-fowler (approved):
Looks good

PR #688, opened 2025-03-12
Update HTTP2StreamChannel.swift

JoachimM7:
Send the whole body, not only a part... otherwise the connection will be lost, when sending big data.

### DIFF
--- a/Sources/HummingbirdHTTP2/HTTP2StreamChannel.swift
+++ b/Sources/HummingbirdHTTP2/HTTP2StreamChannel.swift
@@ -68,15 +68,45 @@ struct HTTP2StreamChannel: ServerChildChannel {
 
                     // read first part, verify it is a head
                     guard let part = try await iterator.next() else { return }
-                    guard case .head(let head) = part else {
+                    guard case .head(var head) = part else {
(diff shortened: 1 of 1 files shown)


adam-fowler (commented):
Not sure what you are fixing here. It seems like you have copied the HTTP1 code across which needs to deal with the possibility of multiple request occurs in sequence on the same connection. A HTTP2 stream channel only ever has to deal with one request. 

The idea that we might need to read the whole request is interesting though. 

I added a `while try await iterator.next() != nil {}` after the calling `responder()` and this fixes your issue to some extent. But it seems to cause issues further down the line as I am seeing errors later on.

JoachimM7:
Thanks for looking at this. Yes, I copied the code from HTTP1, because I realized, that when requesting big data with HTTP1, this code will be called while when using HTTP2, there is only one write and then `self._outbound.finish()`. So somehow the `NIOAsyncChannelOutboundWriter` finishes without writing all data while when using HTTP1, all the data is sent to the client. It's not a SwiftNIO-problem, as the close of the socket is called when the `suspendedYields` of the `NIOAsyncWriter` is empty (although there are more data!)

adam-fowler:
Can I close this now?

PR #689, opened 2025-03-13
Temp fix: wait until the client ends HTTP2 stream

adam-fowler:
Fix for HTTP2 stream writes not getting written. This is a temporary fix until SwiftNIO fix the issue where calls to `NIOAsyncChannelOutboundWriter.write()` don't get flushed before a channel close.

Initially I thought this was relying on the client to end the stream but it looks like the end stream is being handled internally by NIO so we are fine. When HTTP2ChannelHandler receives the `.end` from the response being written out it ends the inbound stream. I've checked this with safari, chrome, Firefox, curl and wget all seem to work.


Joannis (approved):
Unfortunate that this is needed

PR #697, opened 2025-03-27
Add Conditional request support

adam-fowler:
```swift
router.get("ifNoneMatch") { request, context -> Response in
    // if request header ifNoneMatch does not include 1234 return response generated by closure
    // otherwise return status `.notModified` or `.preconditionFailed` depending on method
    try await request.ifNoneMatch(eTag: "1234", context: context) {
        "Hello"
    }
}
router.get("ifModifiedSince") { request, context in
    // if request has ifModifiedSince header and date in that header is before the supplied date
    // return response generated by closure otherwise return status `.notModified`
    try await request.ifModifiedSince(modificationDate: date, context: context) {
        Response(status: .ok)
    }
}
```

The PR also adds support for `Request.ifMatch` and `Request.ifUnmodifiedSince`


Sources/Hummingbird/Server/Response.swift
```diff
@@ -14,6 +14,12 @@
 
 import HummingbirdCore
 
+#if canImport(FoundationEssentials)
```
Cyberbeni, line 17: leftover imports

Sources/Hummingbird/Server/Request.swift
```diff
@@ -57,6 +63,150 @@ extension Request {
...
+    ///   - process: Closure to run if eTag is not in the `If-None-Match` header
+    /// - Returns: Response
+    public func ifUnmodifiedSince(
+        headers: HTTPFields = [:],
+        modificationDate: Date,
+        context: some RequestContext,
+        process: () async throws -> ResponseGenerator
+    ) async throws -> Response {
+        var headers = headers
+        headers[.lastModified] = modificationDate.httpHeader
```
Cyberbeni, line 193: Aren't `If-Unmodified-Since` and `If-Match` used for making changes, creating new eTag/modificationDate, making the current one obsolete?

Sources/Hummingbird/Server/Request.swift
```diff
@@ -57,6 +63,150 @@ extension Request {
...
+        headers[.eTag] = eTag
+        let ifNoneMatch = self.headers[values: .ifNoneMatch]
+        if ifNoneMatch.count > 0 {
+            if ifNoneMatch.contains(eTag) {
+                // Response status based on whether this is a read-only request ie GET or HEAD
+                let status: HTTPResponse.Status =
+                    if self.method == .get || self.method == .head {
+                        .notModified
+                    } else {
+                        .preconditionFailed
```
Cyberbeni, line 94: Based on the documentation, `preconditionFailed` is only for
> This happens with [conditional requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Conditional_requests) on methods other than [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) or [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) when the condition defined by the [If-Unmodified-Since](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Unmodified-Since) or [If-Match](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/If-Match) headers is not fulfilled.

`If-None-Match` is allowed for other methods:

> The HTTP If-None-Match [request header](https://developer.mozilla.org/en-US/docs/Glossary/Request_header) makes a request [conditional](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Conditional_requests). The server returns the requested resource in [GET](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/GET) and [HEAD](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods/HEAD) methods with a [200](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/200) status, only if it doesn't have an [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) matching the ones in the If-None-Match header. For other methods, the request will be processed only if the eventually existing resource's [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/ETag) doesn't match any of the values listed.

If anything, this should be a 500 in this case with an assertion, probably at the start of the function.

Cyberbeni:
Tested `ifModifiedSince` in my project, seems to work as expected.

Sources/Hummingbird/Server/Response.swift
Joannis: Yeah we don't need these

Sources/Hummingbird/Server/Request.swift
adam-fowler: Main uses are to avoid issues, where server side content changes between two requests. eg User read wikis, sends update to that wiki. A 412 is returned is wiki was updated between read and updating to avoid the losing the update. The other case I know of is user is downloading file using ranges. They need to be sure they are downloading from the same file if they send a second request with a different range.

Sources/Hummingbird/Server/Request.swift
```diff
@@ -57,6 +63,150 @@ extension Request {
...
+    ///   - headers: Headers to add to response when condition fails
+    ///   - modificationDate: Modification date for this responses content
+    ///   - context: Request context
+    ///   - process: Closure to run if eTag is not in the `If-None-Match` header
+    /// - Returns: Response
+    public func ifModifiedSince(
+        headers: HTTPFields = [:],
+        modificationDate: Date,
+        context: some RequestContext,
+        process: () async throws -> ResponseGenerator
```
Joannis, line 151: `some`

Sources/Hummingbird/Server/Request.swift
```diff
@@ -57,6 +63,150 @@ extension Request {
...
+    ///   - headers: Headers to add to response when condition fails
+    ///   - eTag: ETag for this responses content
+    ///   - context: Request context
+    ///   - process: Closure to run if eTag is not in the `If-None-Match` header
+    /// - Returns: Response
+    public func ifNoneMatch(
+        headers: HTTPFields = [:],
+        eTag: String,
+        context: some RequestContext,
+        process: () async throws -> ResponseGenerator
```
Joannis, line 82: `some`

Joannis (changes requested):
Some minor tweaks

Joannis (approved):

PR #700, opened 2025-04-17
Add plaintext HTTP2 support

adam-fowler:
Inspired by discussion in https://forums.swift.org/t/swiftnio-http2-server-without-tls/79166/

Realised it'd be really easy to add support. Its quite an obscure case, what do you think?


Joannis (approved):
I'm not opposed. Looks fine by me.

PR #702, opened 2025-04-19
Deal with HTTP parser errors

adam-fowler:
The `HTTPServerProtocolErrorHandler` channel handler sends a badRequest response if the HTTP request parser fails eg in the situation where a http request is not finished. Hummingbird didn't know about this and continued to send its own response. If the pipelining channel handler is running it crashes because we have sent two responses for one request.

The solution is for Hummingbird to not send a response if a request handler throws an HTTP parser error (which they will got from parsing the request body).

This was only caught because pipelining was enabled for WebSocket upgrade channels. I thought about re-enabling it to catch other similar errors but for some reason it breaks the inbound closure stuff, so haven't so far. I've added the ability to set this internally so we can use it in tests

### DIFF
--- a/Tests/HummingbirdTests/ApplicationTests.swift
+++ b/Tests/HummingbirdTests/ApplicationTests.swift
@@ -24,6 +24,7 @@ import Logging
 import NIOConcurrencyHelpers
 import NIOCore
 import NIOEmbedded
+import NIOHTTPTypes
 import NIOSSL
 import ServiceLifecycle
 import Synchronization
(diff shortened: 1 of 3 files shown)


Tests/HummingbirdTests/ApplicationTests.swift
```diff
@@ -1159,6 +1161,70 @@ final class ApplicationTests: XCTestCase {
...
+                    request.body = .stream(stream, length: .known(Int64(4096)))
+                    let response = try await httpClient.execute(request, deadline: .now() + .minutes(30))
+                    let result = try await response.body.collect(upTo: .max)
+                    print("Result size: \(result.readableBytes)")
+                }
+
+                try await Task.sleep(for: .seconds(1))
+                task.cancel()
+                try await Task.sleep(for: .seconds(1))
+                //                _ = try await group.next()
```
Joannis, line 1219: Commented code

Joannis (approved):

Joannis (approved):

PR #705, opened 2025-04-22
Check for multiple responses in testHTTPProtocolParseError

adam-fowler:
Do this by enabling pipeline assistance, which verifies we aren't sending multiple responses


Joannis (approved):
What was CreateErrorHandler used for?

adam-fowler:
> What is CreateErrorHandler used for again?

It's a ChannelHandler that fires errors at inopportune moments.

PR #706, opened 2025-04-22
Clean parse error handling

adam-fowler:
Now that we've moved the parse error handling to Hummingbird there are a number of situations where we were not catching them.

- Moved HTTPParserError handling to HTTPChannelHandler.swift
- Don't throw HTTPParseError if in the middle of writing a response, so we don't start to write another response
- Added a bunch of tests that sends data to the HTTP1Channel directly

### DIFF
--- a/Tests/HummingbirdCoreTests/HTTP1ChannelTests.swift
+++ b/Tests/HummingbirdCoreTests/HTTP1ChannelTests.swift
@@ -37,9 +37,9 @@ final class HTTP1ChannelTests: XCTestCase {
         let logger = Logger(label: "HTTP1Channel")
 
         let http1Channel = HTTP1Channel(responder: responder)
-        let value = try await channel.eventLoop.submit {
+        let value = try await channel.eventLoop.flatSubmit {
             http1Channel.setup(channel: channel, logger: logger)
-        }.get().get()
+        }.get()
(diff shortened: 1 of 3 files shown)


adam-fowler:
API breakage is reported on symbol with package scope 😠

Joannis:
[I've reported it](https://github.com/swiftlang/swift/issues/81003)

Tests/HummingbirdCoreTests/HTTP1ChannelTests.swift
```diff
@@ -0,0 +1,110 @@
...
+            )
+        }
+    ) async throws {
+        let channel = NIOAsyncTestingChannel()
+        let logger = Logger(label: "HTTP1Channel")
+
+        let http1Channel = HTTP1Channel(responder: responder)
+        let value = try await channel.eventLoop.submit {
+            http1Channel.setup(channel: channel, logger: logger)
+        }.get().get()
```
Joannis, line 42:
```suggestion
        let value = try await channel.eventLoop.flatSubmit {
            http1Channel.setup(channel: channel, logger: logger)
        }.get()
```

Joannis (approved):
What an interesting rabbit hole we're getting into.

PR #707, opened 2025-04-27
Add tidy frequency to MemoryPersistDriver

adam-fowler:
- Fix bug where expired test was the wrong way around
- Add test for tidy

### DIFF
--- a/Sources/Hummingbird/Storage/MemoryPersistDriver.swift
+++ b/Sources/Hummingbird/Storage/MemoryPersistDriver.swift
@@ -29,6 +29,20 @@ public actor MemoryPersistDriver<C: Clock>: PersistDriver where C.Duration == Du
         }
     }
 
+    /// Initialize MemoryPersistDriver
+    /// - Parameters:
+    ///   - clock: Clock to use when calculating expiration dates
+    @_disfavoredOverload
+    public init(_ clock: C = .continuous) {
+        self.values = [:]
+        self.clock = clock
+        self.configuration = .init()
+    }
+
+    /// Initialize MemoryPersistDriver
+    /// - Parameters:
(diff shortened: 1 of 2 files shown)


Sources/Hummingbird/Storage/MemoryPersistDriver.swift
```diff
@@ -18,9 +18,35 @@ import NIOCore
 
 /// In memory driver for persist system for storing persistent cross request key/value pairs
 public actor MemoryPersistDriver<C: Clock>: PersistDriver where C.Duration == Duration {
+    public struct Configuration: Sendable {
+        /// amount of time between each call to tidy
+        public var tidyFrequency: Duration
```
Joannis, line 23: Can we bundle the clock in this Configuration too? Since it's within the Generic context of `MemoryPersistDriver`, this type is generic too.

Sources/Hummingbird/Storage/MemoryPersistDriver.swift
```diff
@@ -51,7 +77,7 @@ public actor MemoryPersistDriver<C: Clock>: PersistDriver where C.Duration == Du
         let now = self.clock.now
         self.values = self.values.compactMapValues {
             if let expires = $0.expires {
-                if expires > now {
+                if expires < now {
```
Joannis, line 80: Oops

Sources/Hummingbird/Storage/MemoryPersistDriver.swift
adam-fowler: Can't easily do that and have a default variable for the clock

Joannis: Ah, bummer.

Joannis (approved):

PR #708, opened 2025-04-30
Close on outbound finish

adam-fowler:
Instead of expecting `executeThenClose` to flush writes (which it should) we add the `isOutboundHalfClosureEnabled` to the NIOAsyncChannel and finish the outbound writer when we are done and wait for the channel to close.

I have also added `testWriteLargeBodyAndClose()` which currently fails, but should start working when https://github.com/apple/swift-nio/pull/3148 makes it into a release.


Joannis (approved):
We should make a note not to forget bumping the NIO dependency once that lands

adam-fowler:
> We should make a note not to forget bumping the NIO dependency once that lands

We also need to update hummingbird-websocket as well. It needs the `isOutboundHalfClosureEnabled` set, I can do that now and release straight off.

adam-fowler:
Interesting tests don't fail here

PR #711, opened 2025-05-21
Don't override logLevel when environment variable not found or invalid

### LINKED ISSUE #710: Don't override logLevel when LOG_LEVEL environment is not present
https://github.com/hummingbird-project/hummingbird/blob/9e13ff36994b1650a1b21c1f45aa2a894a8c6da9/Sources/Hummingbird/Application.swift#L230

I am currently writing my own LogHandler and noticed that even though my LogHandler factory returns a handler with higher default logLevel in `LoggingSystem.bootstrap`, I am still getting the info log from `HummingbirdCore/Server.swift:248`

The default LogHandler in swift-log is set to .info, so removing the fallback won't  change anything for most people.

Cyberbeni:
Fixes #710


adam-fowler (approved):
Looks good. Thanks

PR #712, opened 2025-05-22
SendableMetatype fixups

adam-fowler:
Swift 6.2 introduces the protocol `SendableMetatype` for Metatypes that are Sendable. 

Previously all meta types were Sendable but [SE-0470: Global-actor isolated conformances](https://forums.swift.org/t/se-0470-global-actor-isolated-conformances/78704) introduced the concept of isolating conformances. Types that take advantage of this, their metatype cannot be considered Sendable. So `SendableMetatype` was added to enforce Sendability.

Anyway this means certain generic functions require `SendableMetatype` for their generic parameters.

### DIFF
--- a/Sources/Hummingbird/Router/ResponseGenerator.swift
+++ b/Sources/Hummingbird/Router/ResponseGenerator.swift
@@ -13,14 +13,22 @@
 //===----------------------------------------------------------------------===//
 
 import HTTPTypes
+import HummingbirdCore
 
 /// Object that can generate a `Response`.
 ///
 /// This is used by `Router` to convert handler return values into a `Response`.
+#if compiler(>=6.2)
+public protocol ResponseGenerator: SendableMetatype {
+    /// Generate response based on the request this object came from
+    func response(from request: Request, context: some RequestContext) throws -> Response
+}
+#else
 public protocol ResponseGenerator {
     /// Generate response based on the request this object came from
     func response(from request: Request, context: some RequestContext) throws -> Response
 }
+#endif
 
 /// Extend Response to conform to ResponseGenerator
 extension Response: ResponseGenerator {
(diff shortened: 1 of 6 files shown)


Sources/Hummingbird/Router/ResponseGenerator.swift
```diff
@@ -13,14 +13,22 @@
 //===----------------------------------------------------------------------===//
 
 import HTTPTypes
+import HummingbirdCore
 
 /// Object that can generate a `Response`.
 ///
 /// This is used by `Router` to convert handler return values into a `Response`.
+#if compiler(>=6.2)
+public protocol ResponseGenerator: SendableMetatype {
```
Joannis, line 22:
```suggestion
public protocol ResponseGenerator: _HB_SendableMetatype{
```

Shouldn't this use the typealias/protocol like the rest?

adam-fowler: You can't do that because it is used in conditional conformances. eg `extension Optional: ResponseGenerator where Wrapped: ResponseGenerator {}` and in those cases it then pre 6.2 it expects me to include the SendableMetatype conformance but 6.2 complains about an unnecessary repeated conformance if I do include it.

Sources/Hummingbird/Router/ResponseGenerator.swift
```diff
@@ -18,17 +18,11 @@ import HummingbirdCore
 /// Object that can generate a `Response`.
 ///
 /// This is used by `Router` to convert handler return values into a `Response`.
-#if compiler(>=6.2)
-public protocol ResponseGenerator: SendableMetatype {
+@preconcurrency
+public protocol ResponseGenerator: _HB_SendableMetatype {
```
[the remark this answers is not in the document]
adam-fowler: As far as I know tagging it as `@preconcurrency` means the requirement is only valid when strict concurrency is enabled. It softens what is in effect a breaking change

PR #715, opened 2025-05-30
Add support for optional arrays and maps to URLEncodedFormDecoder

nicksloan:
Maps and arrays were excluded from unboxNil and it's not clear to me why. This PR handles those cases.

### DIFF
--- a/Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedFormDecoder.swift
+++ b/Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedFormDecoder.swift
@@ -505,7 +505,7 @@ extension _URLEncodedFormDecoder {
         case .array, .map:
             return false
         case .empty:
-            return true
+            return false
         }
     }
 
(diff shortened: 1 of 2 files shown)


adam-fowler (approved):
Looks good, just need the formatting

Sources/Hummingbird/Codable/URLEncodedForm/URLEncodedFormDecoder.swift
```diff
@@ -499,10 +499,14 @@ extension _URLEncodedFormDecoder: SingleValueDecodingContainer {
...
     func unboxNil(_ node: URLEncodedFormNode) throws -> Bool {
-        guard case .leaf(let value) = node else {
-            throw DecodingError.dataCorrupted(.init(codingPath: self.codingPath, debugDescription: "Expect value not array or dictionary"))
+        switch node {
+            case .leaf(let value):
+                return value == nil
+            case .array, .map:
+                return false
+            default:
+                throw DecodingError.dataCorrupted(.init(codingPath: self.codingPath, debugDescription: "Expect value not array or dictionary"))
```
adam-fowler, line 508:
```suggestion
            case .empty:
                return true
```

nicksloan:
@adam-fowler I'll push an update to make it return false, but I'm pretty sure that code is just unreachable. I tried a bunch of different query strings, and couldn't get it to reach the empty case. I think your previous commit that handles as a non-optional empty array or dictionary precludes us from ever reaching the empty case of unboxNil.

adam-fowler (approved):

PR #717, opened 2025-06-18
Make the statemachine non-copyable to avoid array copies

adam-fowler:


Joannis:
Can't we already drop 5.10 in september?

adam-fowler:
Yes that's why I've left this up here.

PR #727, opened 2025-08-12
Add cookie validation APIs and assert (on DEBUG) when initalizing an invalid cookie

### LINKED ISSUE #726: Hummingbird allows setting of illegal cookies that it can't read
## Hummingbird will (silently) set a cookie that Hummingbird can't read

> TL;DR: you can silently use `;` in the value when setting a cookie, and Hummingbird truncates that result when it tries to read it back. Fixing this is murky.

I ran into this today and it took a bunch of digging and head scratching to figure out what happened and what to do about it.

I was setting a cookie to display a message to my (trusted/admin) user, and it contained an erorr message. The error message came from other code (Soto DynamoDB), but I was including it in the cookie for later display.

Unfortunately, and not explicitly known to me, the error message was `ValidationException: Invalid UpdateExpression: Attribute name is a reserved keyword; reserved keyword: null.`

There's some other machinery in place like wrapping that in a JSON envelope, but the main problem is illuatrated with just that string. It contains a `;` and `;` is illegal in cookies. The *illegal* bit might not be so important if Hummingbird itself didn't fail when trying to read this cookie from the request. It was truncated as `ValidationException: Invalid UpdateExpression: Attribute name is a reserved keyword` since the cookie parser splits on `;`. (This use of `;` as a control character is why `;` is illegal.)

### Background

[RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265) defines `cookie-octet`:

```
 cookie-octet      = %x21 / %x23-2B / %x2D-3A / %x3C-5B / %x5D-7E
                       ; US-ASCII characters excluding CTLs,
                       ; whitespace DQUOTE, comma, semicolon,
                       ; and backslash
```

There are a bunch of restrictions on how cookies are named, but in this context, this is where we are forbidden from using `;` and other characters. (Note: `CTLs` is defined in another RFC, [5234](https://datatracker.ietf.org/doc/html/rfc5234#appendix-B.1), as `%x00-1F / %x7F`).

### Difficulty

There is not a straightforward fix to this problem, in my opinion. Especially without breaking backward compatibility of Hummingbird, existing cookies, or interactions with other systems.

RFC 6265 suggests:

> To maximize compatibility with user agents, servers that wish to store arbitrary data in a cookie-value SHOULD encode that data, example, using Base64

This is how I worked around this issue in my application code—I started encoding these error messages with base64, opting to ignore the loss of backward compatibility in my own cookies, since they're intended to be short-lived anyway.

The hard parts are:
- if Hummingbird tries to do its own implicit encoding (such as with Base64, or as some other platforms do: url encoding), it makes existing cookies impossible or at least much more difficult to use (it would have to *attempt* to decode and fall back to no-encoding or something along those lines)
- if the cookies that Hummingbird sets are intended for another system, implicitly encoding could make it difficult or impossible for the other system to decode the cookie (the encoding would have to be negotiated outseif of the HTTP requests).

#### Solution

I think the best approach here is to make the app handle the encoding. No changes are needed to Hummingbird to allow the app to encode and decode cookies (the coding would happen outside of Hummingbird).

*However*, I do think it's a problem that Hummingbird allowed me to set a cookie that it couldn't read, without as much as a warning (I know it doesn't do this intentionally; the `;` split is a side effect of a mostly-naïve splitting parser). I would really have appreciated a code-focused prevention mechanism. Ideally, I think the `Cookie` `init(…)` would throw an error if there are illegal characters in the name or value, on `init`. But, currently `init` is not marked `throws`, and I don't know of a way to deprecate the current `init` and add a throwing `init` without breaking existing apps.

A cookie that was set to an illegal value *today* might only be read (incorrectly) *far* into the future. Any prevention we can do at development time would be very helpful.

### Patch (Test)

Here's a patch for a test case to illustrate this.

```diff

diff --git a/Tests/HummingbirdTests/CookiesTests.swift b/Tests/HummingbirdTests/CookiesTests.swift
index d17959d..b912d78 100644
--- a/Tests/HummingbirdTests/CookiesTests.swift
+++ b/Tests/HummingbirdTests/CookiesTests.swift
@@ -105,4 +105,26 @@ final class CookieTests: XCTestCase {
             }
         }
     }
+
+    func testIllegalCookie() async throws {
+        let illegalCookieValue = "one ; two"
+        let router = Router()
+        router.post("/") { _, _ -> Response in
+            var response = Response(status: .ok, headers: [:], body: .init())
+            response.setCookie(.init(name: "test", value: illegalCookieValue))
+            return response
+        }
+        let app = Application(responder: router.buildResponder())
+        try await app.test(.router) { client in
+            try await client.execute(uri: "/", method: .post) { response in
+                guard let setCookie = response.headers[.setCookie] else {
+                    XCTFail("No Set-Cookie header found")
+                    return
+                }
+                let cookie = Cookie(from: setCookie[...])
+                XCTAssertEqual(cookie?.value, illegalCookieValue)
+            }
+        }
+    }
+
 }
```

#### Test run

```
❯ swift test --filter CookieTests.testIllegalCookie
[1/1] Planning build
Building for debugging...
[1/1] Write swift-version-12C291108A3FC14E.txt
Build complete! (4.97s)
Test Suite 'Selected tests' started at 2025-08-11 21:50:44.148.
Test Suite 'hummingbirdPackageTests.xctest' started at 2025-08-11 21:50:44.149.
Test Suite 'CookieTests' started at 2025-08-11 21:50:44.149.
Test Case '-[HummingbirdTests.CookieTests testIllegalCookie]' started.
/Users/sean/src/hummingbird/Tests/HummingbirdTests/CookiesTests.swift:125: error: -[HummingbirdTests.CookieTests testIllegalCookie] : XCTAssertEqual failed: ("Optional("one ")") is not equal to ("Optional("one ; two")")
Test Case '-[HummingbirdTests.CookieTests testIllegalCookie]' failed (0.090 seconds).
Test Suite 'CookieTests' failed at 2025-08-11 21:50:44.239.
         Executed 1 test, with 1 failure (0 unexpected) in 0.090 (0.090) seconds
Test Suite 'hummingbirdPackageTests.xctest' failed at 2025-08-11 21:50:44.239.
         Executed 1 test, with 1 failure (0 unexpected) in 0.090 (0.090) seconds
Test Suite 'Selected tests' failed at 2025-08-11 21:50:44.239.
         Executed 1 test, with 1 failure (0 unexpected) in 0.090 (0.092) seconds
◇ Test run started.
↳ Testing Library Version: 6.1 (43b6f88e2f2712e)
↳ Target Platform: arm64-apple-macosx
✔ Test run with 0 tests passed after 0.001 seconds.
```

Joannis:
Fixes #726

### DIFF
--- a/Sources/Hummingbird/HTTP/Cookie.swift
+++ b/Sources/Hummingbird/HTTP/Cookie.swift
@@ -102,48 +102,11 @@ public struct Cookie: Sendable, CustomStringConvertible {
             (byte >= 0x21 && byte != 127 && !separators.contains(byte))
         }
     }
-    
-    /// Create `Cookie`
-    /// - Parameters:
-    ///   - name: Name of cookie
-    ///   - value: Value of cookie
-    ///   - expires: indicates the maximum lifetime of the cookie
-    ///   - maxAge: indicates the maximum lifetime of the cookie in seconds. Max age has precedence over expires (not all user agents support max-age)
-    ///   - domain: specifies those hosts to which the cookie will be sent
-    ///   - path: The scope of each cookie is limited to a set of paths, controlled by the Path attribute
-    ///   - secure: The Secure attribute limits the scope of the cookie to "secure" channels
-    ///   - httpOnly: The HttpOnly attribute limits the scope of the cookie to HTTP requests
-    public static func validated(
-        name: String,
-        value: String,
-        expires: Date? = nil,
-        maxAge: Int? = nil,
-        domain: String? = nil,
-        path: String? = nil,
-        secure: Bool = false,
-        httpOnly: Bool = true
-    ) throws -> Cookie {
-        guard Cookie.isValidName(name) else {
-            throw ValidationError(reason: .invalidName)
-        }
-        
-        guard Cookie.isValidValue(value) else {
-            throw ValidationError(reason: .invalidValue)
-        }
-        
-        return Cookie(
-            name: name,
-            value: value,
-            expires: expires,
-            maxAge: maxAge,
-            domain: domain,
-            path: path,
-            secure: secure,
-            httpOnly: httpOnly
-        )
-    }
 
-    /// Create `Cookie`
+    /// Create `Cookie` and validates the name and value to be valid as per RFC 6265.
+    ///
+    /// If the name and value are not valid, an `ValidationError` will be thrown. Contrary to the equivalent initializer, this function will not `assert` on DEBUG for invalid names and values.
+    ///
     /// - Parameters:
     ///   - name: Name of cookie
     ///   - value: Value of cookie
@@ -163,7 +126,7 @@ public struct Cookie: Sendable, CustomStringConvertible {
         path: String? = nil,
         secure: Bool = false,
         httpOnly: Bool = true,
-        sameSite: SameSite
+        sameSite: SameSite? = nil
     ) throws -> Cookie {
         guard Cookie.isValidName(name) else {
             throw ValidationError(reason: .invalidName)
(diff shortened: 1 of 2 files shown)


adam-fowler (commented):
In general looks good

Sources/Hummingbird/HTTP/Cookie.swift
```diff
@@ -51,6 +60,133 @@ public struct Cookie: Sendable, CustomStringConvertible {
...
+            UInt8(ascii: "}"),
+        ]
+        // Space is no in the separators, but is added to the CTLs because it's right behind the last CTL
+        // `0x1F` is the last CTL, and space is `0x20`
+        return !name.isEmpty && name.utf8.allSatisfy { byte in
+            (byte >= 0x21 && byte != 127 && !separators.contains(byte))
+        }
+    }
+    
+    /// Create `Cookie`
```
adam-fowler, line 106: Can we extend the comment here to indicate the difference between this and the standard `init`. Also can we add a comment to the `init` to point people to the `validated` version if they want validation.

PR #728, opened 2025-08-13
Add cookie validation APIs and deprecate current init(s)

### LINKED ISSUE #726: Hummingbird allows setting of illegal cookies that it can't read

scoates:
An alternate proposal for #727
Fixes #726

@Joannis I humbly suggest this alternative to your PR ; I think it addresses my concerns. Very open to discuss, though.

### DIFF
--- a/Sources/Hummingbird/HTTP/Cookie.swift
+++ b/Sources/Hummingbird/HTTP/Cookie.swift
@@ -25,10 +25,10 @@ public struct Cookie: Sendable, CustomStringConvertible {
             case invalidName
             case invalidValue
         }
-        
+
         let reason: Reason
     }
-    
+
     public enum SameSite: String, Sendable {
         case lax = "Lax"
         case strict = "Strict"
@@ -60,7 +60,11 @@ public struct Cookie: Sendable, CustomStringConvertible {
     public var httpOnly: Bool { self.properties[.httpOnly] != nil }
     /// The SameSite attribute lets servers specify whether/when cookies are sent with cross-origin requests
     public var sameSite: SameSite? { self.properties[.sameSite].map { SameSite(rawValue: $0) } ?? nil }
-    
+    /// indicate if a cookie has been validated
+    public var valid: Bool {
+        Cookie.isValidName(self.name) && Cookie.isValidValue(self.value)
+    }
+
     private static func isValidValue(_ value: String) -> Bool {
         // RFC 6265 Section 4.1.1: cookie-octet set
         // Allowed: 0x21, 0x23-0x2B, 0x2D-0x3A, 0x3C-0x5B, 0x5D-0x7E
@@ -72,7 +76,7 @@ public struct Cookie: Sendable, CustomStringConvertible {
             (0x5D...0x7E).contains(byte)
         }
     }
-    
+
     private static func isValidName(_ name: String) -> Bool {
         // RFC 2616 Section 2.2: token = 1*<any CHAR except CTLs or separators>
         // CTLs: 0-31, 127
@@ -102,46 +106,6 @@ public struct Cookie: Sendable, CustomStringConvertible {
             (byte >= 0x21 && byte != 127 && !separators.contains(byte))
         }
     }
-    
-    /// Create `Cookie`
-    /// - Parameters:
-    ///   - name: Name of cookie
-    ///   - value: Value of cookie
-    ///   - expires: indicates the maximum lifetime of the cookie
-    ///   - maxAge: indicates the maximum lifetime of the cookie in seconds. Max age has precedence over expires (not all user agents support max-age)
-    ///   - domain: specifies those hosts to which the cookie will be sent
-    ///   - path: The scope of each cookie is limited to a set of paths, controlled by the Path attribute
-    ///   - secure: The Secure attribute limits the scope of the cookie to "secure" channels
-    ///   - httpOnly: The HttpOnly attribute limits the scope of the cookie to HTTP requests
-    public static func validated(
-        name: String,
-        value: String,
-        expires: Date? = nil,
-        maxAge: Int? = nil,
-        domain: String? = nil,
-        path: String? = nil,
-        secure: Bool = false,
-        httpOnly: Bool = true
-    ) throws -> Cookie {
-        guard Cookie.isValidName(name) else {
-            throw ValidationError(reason: .invalidName)
-        }
-        
-        guard Cookie.isValidValue(value) else {
-            throw ValidationError(reason: .invalidValue)
-        }
-        
-        return Cookie(
-            name: name,
-            value: value,
-            expires: expires,
-            maxAge: maxAge,
-            domain: domain,
-            path: path,
-            secure: secure,
-            httpOnly: httpOnly
-        )
-    }
 
     /// Create `Cookie`
     /// - Parameters:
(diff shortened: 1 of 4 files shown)


nicksloan:
But the assertion only happens in debug if the user happens to exercise that code path, which may be unlikely if that code path appeared to be working.

We discovered that this code was broken months after building the feature, purely by accident of the cookie value. There are likely users who have broken code in production right now and have no idea because it is just rare enough for illegal characters to appear in their cookies that this issue flies under the radar.

If the user Base64 encodes their cookies, it is true that they don't need this, but it's also true that this won't hurt. They get a notice to change their code, which they can address at their convenience.

I like the deprecation approach because it is a low cost way to let people know their code may be broken without making them do anything immediately if they know that it isn't. And if they want to clear the message, the entirety of the changes required is to add one Boolean argument to their cookie initializations.

scoates:
FWIW, I don't think deprecating the existing `init` is ideal, either, but Swift won't let us have separate `func x()` and `func x() throws`, so I do think that the best approach here is to hint to the developer that their way of doing things (the old init) is ambiguous.

> There are plenty of places where users don't need the validation. 

Agreed; the validation calls are light, but there is *some* overhead. I even think there are places where users will need to set a cookie that has illegal characters for interoperability reasons (I think that could still be done with string hackery, but that's also extremely error-prone). With this approach, the new correct way to handle a case where you don't want to validate a cookie name or value is to do `try Cookie((same params), validate: false)`.

Similarly, the reason the `validate` parameter doesn't get a default (to `true`, ideally) is to force the user to specify to make the signature include it to call the new `init`s. This can be changed to default `true` when semantic versions allow, but right now it's a compiler limitation, IMO.

As Nick mentioned above, I'd like to be able to catch my ambuity error at development time. The assertions would catch it at runtime *if* I trigger the negative assertion case. This approach pushes the user toward the new init signatures (that invlude `validate: Bool`, and hint that the old way is deprecated.

Inbound cookies are permissive (not validated). There are definitely places in my brief survey of cookies my browser is currently sending where if we rejected *invalid* cookies, we'd reject cookies that are currently working. The cookies can, however, be checked by the app/developer with the new `.valid` computed property. Postel's principle.

Hope my intent a bit more clear with that background info.

PR #747, opened 2025-11-03
Windows support

Joannis:
Adds support to run on Windows platforms.

image

Depends on:
- https://github.com/apple/swift-nio/pull/3433
- https://github.com/apple/swift-nio-extras/pull/294
- https://github.com/swift-server/swift-service-lifecycle/pull/215

## Notes

- Temporarily disables FileSystem I/O **on Windows only**, we should consider moving to NIOFS or fix the code itself. To be decided


adam-fowler (commented):
Most of this is just disabling file operations. I would move to NIOFileSystem but it is still prefixed with an `_` indicating it is not final release and breaking changes maybe included. Although now Vapor have moved that seems unlikely. 

Can you chase NIO guys about state of release and what their recommendation is. I want them to commit to an fixed API before moving.

tib:
I’m also curious about the current state of NIO FS. We’re considering using it in the long run.

adam-fowler:
> I’m also curious about the current state of NIO FS. We’re considering using it in the long run.

As far as I understand NIOFS has never gone to full release because there was an issue with FilePath being defined in the System framework (on macOS) and in the swift-system package. They wanted to resolve issues with that before saying the library wouldn't have any breaking changes. Maybe with FilePath moving to the swift stdlib this will be resolved.

I've given up waiting though and just merged NIOFS.

adam-fowler:
Temporarily using traits to disable sections of code Windows doesn't like. So far we have FileSystemSupport and AsyncHTTPClientSupport. I will remove these before merging. I would be great if we could keep these but that'd be a breaking change.

Package.swift
```diff
@@ -49,12 +52,12 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-http-types.git", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.14.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", from: "2.9.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.100.0"),
-        .package(url: "https://github.com/apple/swift-nio-extras.git", from: "1.34.1"),
+        .package(url: "https://github.com/apple/swift-nio.git", from: "2.101.3"),
+        .package(url: "https://github.com/Joannis/swift-nio-extras.git", branch: "jo/windows-ucrt"),
```
adam-fowler: Cannot merge with this set to a branch

Package.swift
```diff
@@ -37,6 +37,9 @@ let package = Package(
         .executable(name: "PerformanceTest", targets: ["PerformanceTest"]),
     ],
     traits: [
+        .trait(name: "FileSystemSupport", description: "Enable support for file access."),
+        .trait(name: "AsyncHTTPClientSupport", description: "Enable support for async-http-client."),
+        .trait(name: "TLSSupport", description: "Enable support for TLS."),
```
adam-fowler, line 42: Do not merge with these traits

adam-fowler:
Status
- Compiles with SSL, AsyncHTTPClient and FileSystem code disabled by traits
- Tests run with only two failures related to streamed files

Budoman:
Carried this PR downstream to evaluate a Swift server application on Windows, and it forward-ports cleanly. A few data points that may be useful.

**Rebasing onto 2.26.0.** Cherry-picking the 13 commits from the PR head onto tag 2.26.0 produced exactly one conflict: `55bc902` renamed `NIOFileSystem` to `_NIOFileSystem` after this branch's last rebase, so `Sources/Hummingbird/Environment.swift` conflicts between the tag's `import _NIOFileSystem` and this branch's trait-guarded `import NIOFileSystem`. Keeping the release's module name inside the trait guard resolves it; the other file moves merge without conflict.

**swift-nio-extras.** apple/swift-nio-extras#294 merged on 2026-08-06 (`3078359`), so the dependency on the personal fork can go back to upstream. It needs a revision pin for now: the newest tag (1.34.3) predates the merge.

**Measured on windows-2022, Swift 6.3.3.** The carried Hummingbird itself resolves and builds cleanly on Windows with default traits. With `TLSSupport` enabled it fails inside swift-nio-ssl ("unsupported os" plus missing POSIX shims), so TLS on Windows is blocked upstream of this PR, not by it. Our own application code has its own portability work remaining, so this is a statement about Hummingbird, not a full-stack result.

**On the `FileSystemSupport` default.** For at least one real consumer, defaulting it off costs nothing: we use none of FileMiddleware/FileIO/LocalFileSystem, so `TLSSupport` alone was sufficient, and it kept `_NIOFileSystem`, async-http-client and swift-configuration out of the build entirely. One vote that the trait granularity is right.

Happy to share the CI setup if useful.

adam-fowler:
> Carried this PR downstream to evaluate a Swift server application on Windows, and it forward-ports cleanly. A few data points that may be useful.

Thanks for this. It's great to hear you got to the point where it is usable. I'll update the PR to latest from main. As you've noticed most of the code is now reliant on fixes further upstream.

So far my way of testing has been github CI. I need to set myself up with a Windows machine. We have two tests failing in CI. These need diagnosing. Also I have no idea what performance is like.

EDIT: We are now using swift-nio-extras main branch and we have rebased with main branch of hummingbird. So this PR is up to date with latest hummingbird

Budoman:
Re-ran our downstream Windows CI against the rebased head (f84a819) within a day of the rebase. Results and the CI setup, since you mentioned not having a Windows machine handy.

**The rebased head builds clean on Windows.** windows-2022, Swift 6.3.3, default traits: resolve and build both pass with the branch consumed as a dependency, no local patches. The swift-nio-extras move to upstream main resolved without issue, and the rebase let us drop the one local carry commit we previously needed. Our fork now mirrors your PR head verbatim.

**TLS remains blocked upstream, unchanged.** With `TLSSupport` enabled, swift-nio-ssl fails with `error: unsupported os` (544 occurrences) plus missing POSIX type shims (`stat`, `size_t`, `in_addr`, 136 each). Same shape as before the rebase, so nothing in this PR regressed; the wall is entirely in swift-nio-ssl.

**CI setup, in case it is useful as a starting point.** We run a plain GitHub-hosted `windows-2022` runner with phases that `continue-on-error` and write a verdict table to the step summary, because per-step conclusions are misleading once any phase is allowed to fail. The Hummingbird-relevant core is small:

```yaml
runs-on: windows-2022
steps:
  - uses: compnerd/gha-setup-swift@main
    with:
      branch: swift-6.3.3-release
      tag: 6.3.3-RELEASE
  - uses: actions/checkout@v4
  - name: Build Hummingbird (default traits)
    shell: pwsh
    run: swift build 2>&1 | Tee-Object build.log
  - name: Summarize distinct errors
    if: always()
    shell: pwsh
    run: |
      Select-String -Path build.log -Pattern 'error:' |
        ForEach-Object { ($_ -replace '^.*?error:', 'error:').Trim() } |
        Group-Object | Sort-Object Count -Descending |
        ForEach-Object { Add-Content $env:GITHUB_STEP_SUMMARY ("{0,4}x  {1}" -f $_.Count, $_.Name) }
```

The distinct-error grouping is the piece we would most recommend: one broken header cascading through 90 files reads as one finding with a count instead of 90 findings, which makes Windows triage tractable from CI alone.

We build the branch as a consumer rather than running your test suite, so no light to shed on the two streamed-file test failures from this run; happy to add a test-suite phase against the PR head if that would help narrow them.

adam-fowler:
I have a Windows docker image running on a Linux box now.
I think the streaming errors will get resolved by this PR on swift-nio https://github.com/apple/swift-nio/pull/3685

Budoman:
Good timing on the Windows Docker image; that removes the machine-access friction.

We pointed our downstream Windows CI at Hummingbird's own test suite this time (windows-2022, Swift 6.3.3, PR head f84a819, default traits), in an environment independent of yours. It runs 404 tests across 32 suites and reproduces exactly two failures, both in the streamed-body path, which lines up with your count:

- `testStreamBody()` (HummingbirdCore, CoreTests.swift:226): "stream ended at an unexpected time"
- `testConsumeWithCancellationOnInboundClose()` (Hummingbird, ApplicationTests.swift:954): "An existing connection was forcibly closed by the remote host"

Everything else passes. If those are the two you expect apple/swift-nio#3685 to clear, we are set up to re-run with swift-nio pointed at that revision once it is pinnable and report back whether they go green, in an environment separate from yours.

No change on TLS: swift-nio-ssl still hits the unsupported-os wall, upstream of this PR.

adam-fowler:
Unfortunately latest on swift-nio main does not fix the issues with the streaming tests.

Budoman:
Thanks for closing that off; saves everyone chasing a fix that isn't there.

We still reproduce exactly those two on our side (windows-2022, Swift 6.3.3, PR head, 404 tests across 32 suites, only these two red).

One thing worth a look on the second one. "An existing connection was forcibly closed by the remote host" is the literal WSAECONNRESET (10054), which on Windows surfaces in places where POSIX would instead deliver a graceful EOF (a zero-length read). `testConsumeWithCancellationOnInboundClose` rides exactly that path (peer closes, cancel on inbound close), so this may be a Windows close-semantics difference at the transport rather than the streaming path, which would fit main not moving it. That is a hypothesis, not a claim; the WSAECONNRESET mapping is the only solid part.

We are set up to re-run with channel-level tracing around both tests, to capture whether the second is arriving as ECONNRESET vs a clean EOF and whether `testStreamBody` ("stream ended at an unexpected time") is an early half-close on the same seam. Happy to run that and report the raw traces if it would help narrow things.

adam-fowler:
@Budoman checkout https://github.com/apple/swift-nio/issues/3697. If you've anything useful to add to that issue that'd be awesome.

PR #749, opened 2025-11-09
Omit lowercasing components and paths, and opt for case insensitive comparisons instead

Joannis:
UPDATED (AF):
- Use own version of `caseInsensitiveCompare` which means we don't need to import Foundation. It is also faster as `caseInsensitiveCompare` calls another function `compare` which has lots of other options
- Merged in benchmark. This code is slower, but it is correct.
- Got it working for HummigbirdRouter


Sources/Hummingbird/Router/RouterPath.swift
```diff
@@ -12,6 +12,8 @@
 //
 //===----------------------------------------------------------------------===//
 
+internal import Foundation
```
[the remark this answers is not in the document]
adam-fowler: Found it wasn't and wrote own versions

Sources/HummingbirdRouter/Route.swift
```diff
@@ -19,18 +19,16 @@ public struct Route<Handler: _RouteHandlerProtocol, Context: RouterRequestContex
...
 
     /// Initialize Route
     /// - Parameters:
     ///   - method: Route method
     ///   - routerPath: Route path, relative to group route is defined in
     ///   - handler: Route handler
     init(_ method: HTTPRequest.Method, _ routerPath: RouterPath = "", handler: Handler) {
-        let options = RouterBuilderState.current?.options ?? []
-        var routerPath = routerPath
-        if options.contains(.caseInsensitive) {
```
Joannis, line 31: The removal is a performance hit right?

adam-fowler (approved):
Given this is a combination of Joannis' and my code not sure I can approve but whatever

Sources/HummingbirdRouter/Route.swift
adam-fowler: I added the lowercased call back in. There shouldn't be any performance change to have an already lowercased router path. The code path is exactly the same, but it does seem to improve performance.

PR #770, opened 2026-01-09
Use Swift String manipulation API

Cyberbeni:
The current method is [not unicode safe](https://www.hackingwithswift.com/articles/280/one-swift-mistake-everyone-should-stop-making-today), haven't tested performance.

As far as I can tell we'll just need to add a new trait to conditionally exclude DateFormatter support in URLEncodedFormDecoder/Encoder after this to be fully FoundationInternationalization free. (Besides dependencies, for example swift-configuration's default traits depend on Foundation)


adam-fowler:
Your code will introduce an additional allocation when you create the String. 

While replacingOccurences is unsafe when dealing with double character emoji characters it is perfectly safe for searching for 7-bit ASCII characters.

adam-fowler:
Yeah looks like you're correct there, but you shouldn't need to create a `String`. The function `replacing` is a protocol requirement of `RangeReplaceableCollection` and Substring conforms to  `RangeReplaceableCollection`.

PR #788, opened 2026-02-22
Add 6.3 nightly

adam-fowler:
Fix new warning

`Cannot use default expression for inference of 'C' because it is inferrable from parameters #0, #1; this will be an error in a future Swift language mode`


Joannis (approved):
Bit dumb if you ask me, but okay

PR #806, opened 2026-04-08
Add availability macro and extend swift settings to all modules

adam-fowler:
Add `@available(hummingbird 2.0, *)`

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -4,7 +4,7 @@
 import PackageDescription
 
 let swiftSettings: [SwiftSetting] = [
-    .enableExperimentalFeature("AvailabilityMacro=hummingbird 2.0:macOS 14.0, iOS 17.0, tvOS 17.0, visionOS 1.0, android 28"),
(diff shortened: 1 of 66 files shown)


Package.swift
```diff
@@ -1,9 +1,11 @@
-// swift-tools-version:6.2
+// swift-tools-version:6.1
 // The swift-tools-version declares the minimum version of Swift required to build this package.
 
 import PackageDescription
 
 let swiftSettings: [SwiftSetting] = [
+    .enableExperimentalFeature("AvailabilityMacro=hummingbird 2.0:macOS 14.0, iOS 17.0, tvOS 17.0, visionOS 1.0"),
```
Joannis: watchOS or Android annotations needed?

adam-fowler: Does android have an availability version ? WatchOS is pointless as you can't run a server on a watch. You have to jump through hoops to even get a network connection.

Joannis: It does

Joannis: Android I mean

Joannis (approved):

Package.swift
adam-fowler: What would the platform version be for Android?

PR #813, opened 2026-04-15
Add LoggingConnections available connection delegate

adam-fowler:
For logging of when connections are opened and closed. Also added static functions for AvailableConnectionsDelegates for ease of finding

```swift
var app = Application(
    router: router,
    configuration: .init(
        address: .hostname(hostname, port: port),
        availableConnectionsDelegate: .logging(Logger(label: "Connections"), logLevel: .info)
    )
)
```
or for setting the maximum number of connections allowed
```swift
var app = Application(
    router: router,
    configuration: .init(
        address: .hostname(hostname, port: port),
        availableConnectionsDelegate: .maximum(1024)
    )
)
```


Sources/HummingbirdCore/Server/AvailableConnectionsHandler.swift
```diff
@@ -49,6 +50,46 @@ public struct MaximumAvailableConnections: AvailableConnectionsDelegate {
...
+
+    public init(_ logger: Logger, logLevel: Logger.Level = .debug) {
+        self.logger = logger
+        self.logLevel = logLevel
+        self.connectionCount = 0
+    }
+
+    public mutating func connectionOpened() {
+        self.connectionCount += 1
+        self.logger.log(level: logLevel, "Connection opened.", metadata: ["NumConnections": .stringConvertible(self.connectionCount)])
```
Joannis: `NumConnections` is not a canonically formatted OTel string right?

adam-fowler: Probably not, does it need to be? This is just a debug tool, to catch connection leaks. You'd probably not include it in a production server. The production server would include the version that sets the maximum number of connections, to avoid overloading a server.

adam-fowler:
Decided to merge the two connection delegates into one.

Sources/HummingbirdCore/Server/AvailableConnectionsHandler.swift
Joannis: Well, even if you intend it for debug only - people will. And I'd strive to be correct regardless.

Joannis (approved):

PR #814, opened 2026-04-15
Don't set SO_REUSEADDR on child channel

adam-fowler:
It does nothing


Sources/HummingbirdCore/Server/Server.swift
```diff
@@ -247,7 +247,7 @@ public actor Server<ChildChannel: ServerChildChannel>: Service {
             case .unixDomainSocket(let path):
                 let asyncChannel = try await bootstrap.bind(
                     unixDomainSocketPath: path,
-                    cleanupExistingSocketFile: false,
+                    cleanupExistingSocketFile: true,
```
Joannis, line 250: What's this change? It's not mentioned in your PR

adam-fowler: Oh did that leak into the PR. It’s basically a clean up thing to ensure we don’t connect to an existing socket.

Joannis (approved):

PR #818, opened 2026-04-21
Use NIOFounationEssentialsCompat instead of vendored code

Cyberbeni:


Package.swift
```diff
@@ -49,7 +49,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-http-types.git", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-log.git", from: "1.11.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", from: "2.5.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.83.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", revision: "38000794f29a24e11a6f5ea41ad28facb7a255d1"),
```
Joannis: Note: We can't do this of course

Package@swift-6.1.swift
```diff
@@ -37,7 +37,7 @@ let package = Package(
         .package(url: "https://github.com/apple/swift-http-types.git", from: "1.0.0"),
         .package(url: "https://github.com/apple/swift-metrics.git", from: "2.5.0"),
         .package(url: "https://github.com/apple/swift-distributed-tracing.git", from: "1.3.0"),
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.83.0"),
+        .package(url: "https://github.com/apple/swift-nio.git", revision: "38000794f29a24e11a6f5ea41ad28facb7a255d1"),
```
Joannis: Same here

Sources/Hummingbird/Codable/JSON/JSONCoding.swift
```diff
@@ -6,6 +6,8 @@
 // SPDX-License-Identifier: Apache-2.0
 //
 
+private import NIOFoundationEssentialsCompat
```
Joannis, line 9: Very cool! Is this new?

Package.swift
Cyberbeni: This change just got merged into swift-nio, there is no released version yet, that's why the PR is still a draft.

Joannis (approved):

Joannis:
Breaking changes, but I'm not convinced they're true

adam-fowler (approved):

PR #820, opened 2026-04-27
Fix flakiness in testTracingSpanLength

aryan-25:
The `testTracingSpanLength` test sets up a route handler whose response body sleeps for exactly 100 milliseconds, then asserts the recorded span duration is greater than 100 milliseconds.

The assertion uses a strict greater-than check (`> 100`), which fails when the duration lands on exactly 100ms. This is resulting in test runs occasionally failing. A duration of exactly 100ms is a valid result and the test should pass. This change simply updates the assertion condition from `> 100` to `>= 100`.

### DIFF
--- a/Tests/HummingbirdTests/TracingTests.swift
+++ b/Tests/HummingbirdTests/TracingTests.swift
@@ -542,7 +542,7 @@ struct TracingTests {
             }
             let span = try #require(Self.testTracer.spans.first)
             // Test tracer records span times in milliseconds
-            #expect(span.endTime! - span.startTime > 100)
+            #expect(span.endTime! - span.startTime >= 100)
         }
     }
 


Tests/HummingbirdTests/TracingTests.swift
```diff
@@ -542,7 +542,7 @@ struct TracingTests {
             }
             let span = try #require(Self.testTracer.spans.first)
             // Test tracer records span times in milliseconds
-            #expect(span.endTime! - span.startTime > 100)
+            #expect(span.endTime! - span.startTime >= 100)
         }
     }
```
Joannis, line 547: Surprised this ever was an issue..

Joannis (approved):
Great find

Tests/HummingbirdTests/TracingTests.swift
adam-fowler: I must say this is not a test I've seen fail. Are you seeing this fail @aryan-25

aryan-25: Yes, I'm able to hit this case consistently when running the test repeatedly using [Xcode's run test repeatedly feature](https://developer.apple.com/documentation/xcode/running-tests-and-interpreting-results#Run-testing-repeatedly-to-determine-reliability) within 100 repetitions.

adam-fowler (approved):

adam-fowler:
Thanks @aryan-25 it is now merged

Issue #812, opened 2026-04-15
Server hangs after logging `Waiting on child channel`


andreasley:
**Environment:**
hummingbird 2.22.0
swift-nio 2.97.1
Running on macOS 26.4, exposed to public internet (so various shenanigans)

**Problem:**
Service stops accepting connections every few hours, right after the following log entry:
```
Waiting on child channel: setsockopt(socket:level:optionName:optionValue:optionLen:): Invalid argument) (errno: 22)
```

Unfortunately, I can't reproduce this exact problem yet.

Questions:
1. I don't understand why such an error throws all the way to the input stream handler ([possibly related](https://github.com/apple/swift-nio/issues/3479)).

2. If the inbound stream fails, the service shouldn't hang ([possibly related](https://github.com/apple/swift-nio/issues/3472)). Even termination would be preferable to hanging, because then the service can be restarted. Would it be appropriate to set `self.state = .shutdown` [here](https://github.com/hummingbird-project/hummingbird/blob/0435b066d883f2e8a4e8abe10e0d33cff3091877/Sources/HummingbirdCore/Server/Server.swift#L127)?

3. The only `setsockopt` call related to child connections seems to be SO_REUSEADDR, originating [here](https://github.com/hummingbird-project/hummingbird/blob/0435b066d883f2e8a4e8abe10e0d33cff3091877/Sources/HummingbirdCore/Server/Server.swift#L293). Does ist even make sense to set SO_REUSEADDR on child connections?

adam-fowler:
This is a macOS server?
Setting SO_REUSEADDR on the child channel should be fine, have you tried removing that call just in case? 
I'm wondering if there is a chance you are hanging onto connections somehow? You could setup a `AvailableConnectionsDelegate` see `ApplicationConfiguration`. Create a type conforming to `AvailableConnectionsDelegate` and you can print out how many connections you have open.

adam-fowler:
Have you set the number of connections the Mac can open? By default applications on Mac are only allowed 256.

andreasley:
Yes, a macOS server. Rare, I know. :)
Unfortunately, it's not related to the number of connections. 

Once the `executeThenClose` throws (for whatever reason), the channel input stream is broken forever. A single (weird) connection attempt can trigger this. I think it's a SwiftNIO issue that this throw even happens in situations where just a single connection is broken, but I also think that hummingbird could handle the broken channel input stream better (e.g. by shutting down).

[...]

PR #822, opened 2026-04-30
Add ServerErrorHandler to catch errors when creating child channel

adam-fowler:
See #812 and https://github.com/apple/swift-nio/issues/3576


Joannis (approved):

wibed:
has this been benched? i am confuse does it accept the erronuous behaviour and just throw?

adam-fowler:
> has this been benched? i am confuse does it accept the erronuous behaviour and just throw?

It gets around an issue where a closing a connection immediately after opening would report an error on the server channel on macOS. If unhandled it would close the server channel and thus bring down the server. There is nothing we can do with this error so we just log it. 

Read the linked issue.

PR #829, opened 2026-05-13
Use commit hashes for non github owned actions

adam-fowler:


Joannis (approved):
Shouldn't this happen on a template level?

adam-fowler:
@Joannis These two actions are specific to hummingbird. I was going to do the template ones when we had other things to add to the template, to avoid the time spent approving and merging 20 PRs

PR #835, opened 2026-05-19
Security middleware

adam-fowler:
Middleware for setting various resource access related headers. `content-security-policy`, `cross-origin-resource-policy`, `x-content-type-options` and optionally setting `content-security-policy-report-only` and `reporting-endpoints`

I think it needs another name though. `SecurityMiddleware` is very vague. But `ResourcesAccessControlMiddleware` or something similar is quite a mouthful and not much more descriptive. Perhaps add an `HTML` or browser prefix as it is related to web browsers.

### DIFF
--- a/Sources/Hummingbird/Middleware/SecurityMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/SecurityMiddleware.swift
@@ -11,7 +11,7 @@ import HTTPTypes
 /// Middleware for setting up various security related response headers
 ///
 /// Currently sets headers `content-security-policy`, `cross-origin-resource-policy`, `x-content-type-options`
-/// and optionally sets `reporting-endpoints`.
+/// and optionally sets `content-security-policy-report-only` and `reporting-endpoints`.
 public struct SecurityMiddleware<Context: RequestContext>: RouterMiddleware {
     let headers: HTTPFields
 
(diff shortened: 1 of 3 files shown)


Joannis (approved):

Sources/Hummingbird/Middleware/SecurityMiddleware.swift
```diff
@@ -0,0 +1,81 @@
...
+// SPDX-License-Identifier: Apache-2.0
+//
+
+import HTTPTypes
+
+/// Middleware for setting up various security related response headers
+///
+/// Currently sets headers `content-security-policy`, `cross-origin-resource-policy`, `x-content-type-options`
+/// and optionally sets `content-security-policy-report-only` and `reporting-endpoints`.
+public struct SecurityMiddleware<Context: RequestContext>: RouterMiddleware {
```
Joannis: I think ContentSecurityMiddleware makes more sense maybe?

Joannis: Depends on if we intend to make it a catch-all middleware or still applied one-by-one

adam-fowler: `BrowserSecurityMiddleware`? As these only make sense when served to a web browser. `ContentSecurityMiddleware` is also a possibility.

adam-fowler: Gone with ContentSecurityMiddleware

PR #842, opened 2026-05-27
Add PersistDriver.getWithTTL

adam-fowler:
As this is a protocol requirement, I have added a default version as well


thoven87 (approved):
looks good to me. I suppose this could be default in HB 3.0?

PR #862, opened 2026-07-13
Add case insensitive benchmark

adam-fowler:


.github/workflows/ci.yml
```diff
@@ -39,8 +39,8 @@ jobs:
         image: ["swift:6.1", "swift:6.2", "swift:6.3", "swiftlang/swift:nightly-6.4.x-noble"]
     container:
       image: ${{ matrix.image }}
-    env:
-      ENABLE_HB_BENCHMARKS: true
+    #env:
+    #  ENABLE_HB_BENCHMARKS: true
```
Joannis, line 43: What's this change?

Benchmarks/HummingbirdBenchmarks/RouterBenchmarks.swift
```diff
@@ -179,6 +179,18 @@ func routerBenchmarks() {
         return router
     }
 
+    Benchmark(
+        "Router:CaseInsensitive",
+        configuration: .init(warmupIterations: 10),
+        request: .init(method: .get, scheme: "http", authority: "localhost", path: "/test/case/Insensitive/performance")
+    ) {
+        let router = Router(context: BasicBenchmarkContext.self, options: .caseInsensitive)
+        router.get("/test/case/Insensitive/performance") { _, _ in
```
Joannis: Shouldn't we add a bit more weird casing differences in here? This is byte-by-byte equivalent.

.github/workflows/ci.yml
adam-fowler: The benchmark package is broken on swift 6.4. This is a temporary fix to get build CI working. The actual benchmark action runs on swift 6.3 and still runs.

Benchmarks/HummingbirdBenchmarks/RouterBenchmarks.swift
adam-fowler: Fair point

Joannis (approved):

PR #869, opened 2026-07-21
Add serveOnNotFoundResponse flag to FileMiddleware (#592)

### LINKED ISSUE #592: FileMiddleware and 404 responses
Currently the FileMiddleware runs whenever the router throws a 404 HTTPError. It does not run if a Response is returned whose status is 404. 

Should we consider a returned 404 responses as a reason to kick off the FileMiddleware? On one hand a returned response could be considered as already handled and therefore we shouldn't kick off the FileMiddleware. On the other people have been caught out by this. 

Options:
- Treat 404 responses the same as errors. 
- Add an option to treat 404 responses the same as errors. 
- Improve documentation. 

@tib

rogerneal:
This adds an opt-in serveOnNotFoundResponse: Bool = false parameter to FileMiddleware.

Currently, FileMiddleware only serves static files when a downstream handler throws HTTPError.notFound. If a route returns a 404 response directly instead of throwing an error, static file serving is skipped.

Setting serveOnNotFoundResponse: true allows FileMiddleware to catch returned 404 responses and attempt to serve a matching static file. The default remains false to preserve existing behavior.

### Changes
* Added serveOnNotFoundResponse to FileMiddleware initializer and handler configuration.
* Added unit tests covering returned 404 responses, default behavior, HEAD requests, and urlBasePath interactions.

Resolves #592

### DIFF
--- a/Sources/Hummingbird/Middleware/FileMiddleware.swift
+++ b/Sources/Hummingbird/Middleware/FileMiddleware.swift
@@ -45,6 +45,7 @@ public struct FileMiddleware<Context: RequestContext, Provider: FileProvider>: R
 where Provider.FileAttributes: FileMiddlewareFileAttributes {
     let cacheControl: CacheControl
     let searchForIndexHtml: Bool
+    let serveOnNotFoundResponse: Bool
     let urlBasePath: String?
     let fileProvider: Provider
     let mediaTypeFileExtensionMap: [MediaType.FileExtension: MediaType]
@@ -74,6 +75,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
             urlBasePath: urlBasePath,
             cacheControl: cacheControl,
             searchForIndexHtml: searchForIndexHtml,
+            serveOnNotFoundResponse: false,
             mediaTypeFileExtensionMap: [:]
         )
     }
@@ -95,6 +97,7 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
             urlBasePath: urlBasePath,
             cacheControl: cacheControl,
             searchForIndexHtml: searchForIndexHtml,
+            serveOnNotFoundResponse: false,
             mediaTypeFileExtensionMap: [:]
         )
     }
@@ -104,10 +107,12 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
         urlBasePath: String? = nil,
         cacheControl: CacheControl = .init([]),
         searchForIndexHtml: Bool = false,
+        serveOnNotFoundResponse: Bool = false,
         mediaTypeFileExtensionMap: [MediaType.FileExtension: MediaType]
     ) {
         self.cacheControl = cacheControl
         self.searchForIndexHtml = searchForIndexHtml
+        self.serveOnNotFoundResponse = serveOnNotFoundResponse
         self.urlBasePath = urlBasePath.map { String($0.dropSuffix("/")) }
         self.fileProvider = fileProvider
         self.mediaTypeFileExtensionMap = mediaTypeFileExtensionMap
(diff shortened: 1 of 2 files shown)


Joannis:
I don't know if we need a setting like this. Just feels like a bug in our middleware.

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -140,84 +146,112 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
     public func handle(_ request: Request, context: Context, next: (Request, Context) async throws -> Response) async throws -> Response {
         do {
-            return try await next(request, context)
+            let response = try await next(request, context)
+            if self.serveOnNotFoundResponse, response.status == .notFound {
+                do {
+                    return try await self.serveFile(for: request, context: context, fallbackError: HTTPError(.notFound))
+                } catch {
+                    if let httpError = error as? any HTTPResponseError, httpError.status == .notFound {
+                        return response
```
adam-fowler: Should we be returning the original response here? It may have a body with a completely different error message. I think we should probably throw the error thrown by `serveFile`. I ready to be convinced of the opposite though

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -140,84 +146,112 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
+                    }
+                    throw error
+                }
+            }
+            return response
         } catch {
-            // Guard that error is HTTP error notFound
```
adam-fowler, line 152: Why was the comment removed?

Sources/Hummingbird/Middleware/FileMiddleware.swift
```diff
@@ -140,84 +146,112 @@ where Provider.FileAttributes: FileMiddlewareFileAttributes {
...
+            return try await self.serveFile(for: request, context: context, fallbackError: error)
+        }
+    }
 
-            guard request.method == .get || request.method == .head else {
-                throw error
-            }
+    private func serveFile(for request: Request, context: Context, fallbackError: any Error) async throws -> Response {
+        guard request.method == .get || request.method == .head else {
+            throw fallbackError
```
adam-fowler: Should this be a fallbackResult and replace `throw fallbackError` with `return try fallbackResult.get()`

adam-fowler (commented):
In general looks good. Just some questions over what to return/throw when files don't exist, request is not a GET/HEAD

adam-fowler:
> I don't know if we need a setting like this. Just feels like a bug in our middleware.

In the related issue #592 you said you wanted an option. I think it should be an option as it'd be a breaking change otherwise.

Sources/Hummingbird/Middleware/FileMiddleware.swift
rogerneal: Returning the original response makes sense when serveOnNotFoundResponse is enabled. If a downstream route returned a custom 404 response (like custom JSON or HTML error headers/body) and no matching static file was found, returning the original response preserves that route's custom 404 output instead of overwriting it with a generic 404 error.

Refactoring serveFile with fallback: Result<Response, Error> cleanly handles this by returning that original Response if no static file matches.

Sources/Hummingbird/Middleware/FileMiddleware.swift
rogerneal: That was unintentional. I can add it back.

Sources/Hummingbird/Middleware/FileMiddleware.swift
rogerneal: Great idea, that makes the fallback logic much cleaner and gets rid of the nested do-catch in handle. I'll refactor serveFile to take a Result<Response, Error> if you want.

adam-fowler: Yes please

adam-fowler (approved):
Looks good

PR #872, opened 2026-07-23
Replace custom HTTP2ServerConnectionManager with NIOHTTP2ServerConnectionManagementHandler

rogerneal:
Replaces the custom HTTP/2 connection management state machine with swift-nio-http2's official `NIOHTTP2ServerConnectionManagementHandler`.

Closes #798.


adam-fowler (commented):
Everything looks fine. Except it doesn't work. When the client closes a connection the server channel is left dangling and it isn't possible to shutdown the server. That is why the tests are not finishing. Hummingbird version responded to input closed event, NIO version doesn't.

I've added an Issue to swift-nio-http2 https://github.com/apple/swift-nio-http2/issues/558

rogerneal:
@adam-fowler - Do you want to add me to that ticket too to keep eyes on it?

adam-fowler:
> @adam-fowler - Do you want to add me to that ticket too to keep eyes on it?

@rogerneal There's a subscribe button on the right hand side you can use to do that.

PR #873, opened 2026-07-23
Ensure services have actually started before running test closure

### LINKED ISSUE #871: RouterTestFramework can trigger graceful shutdown before the ServiceGroup starts, so app.services never run
### Summary

`RouterTestFramework.run` spawns `serviceGroup.run()` as a child task and then calls `triggerGracefulShutdown()` once the test closure returns — with nothing guaranteeing the child ever started. When the parent wins that race, `ServiceGroup` goes `.initial → .finished`, the child's `run()` throws `alreadyFinished` into a task group that discards it, and **none of `app.services` ever run**.

Version: hummingbird 2.25.0, swift-service-lifecycle 2.11.0, Swift 6.1 on Linux.

```swift
// Sources/HummingbirdTesting/RouterTestFramework.swift
group.addTask {
    try await serviceGroup.run()
}
do {
    // run the runBeforeServer processes before we run test closure. Need to do this
    // after we have run the serviceGroup though
    for process in self.processesRunBeforeServerStart {
        try await process()
    }
    let value = try await test(client)
    await serviceGroup.triggerGracefulShutdown()
    return value
```

The existing comment — *"Need to do this after we have run the serviceGroup though"* — states exactly the guarantee that's missing: `addTask` only *schedules* the child, so the prelude and the whole test body can run before `serviceGroup.run()` executes. A `.test(.router)` body is in-memory and typically finishes in well under a millisecond, which makes the parent winning quite likely rather than exotic.

### Why it bites

Services that own resource teardown ("suspend until `gracefulShutdown()`, then release X") silently never release. In our case a `Service` wrapping an `AsyncHTTPClient.HTTPClient` never ran, so the client was never shut down and tripped AsyncHTTPClient's debug `deinit` precondition — SIGILL, killing the whole XCTest process. Because the trap fires wherever that object is later deallocated, the crash appeared in unrelated test suites, which made it look like a random flake for weeks.

Tests still *pass* when this happens: the discarded `alreadyFinished` means the only symptom is the missing side effect.

### Reproduction

A minimal harness using real `app.test(.router)` + a `Service` holding an `HTTPClient`, on Linux (`swift:6.1-noble`) at `--cpus 4`:

- **9 crashes in 40,473 iterations** (~2.2e-4 per test), each with a discriminator confirming the service body never executed.
- Leak positions uniform across the run — the process just dies at the first one.
- **0 in 80,000** once the app's prelude waits for a signal emitted as the service's first statement.
- Non-monotonic in core count: 0/20,000 at `--cpus 2` and 0/20,000 at 11 cores, but reliable at 4.

### Workaround (for anyone hitting this)

Because `processesRunBeforeServerStart` runs on the parent *before* `triggerGracefulShutdown()`, a start gate closes it without touching the framework:

```swift
app.beforeServerStarts { await startGate.wait() }   // signalled as the first line of the service's run()
```

This is also safe in production, where the prelude runs inside `serverService.run()` — a sibling spawned by the same unconditional `addTask` loop — so the signal always arrives. We bound the wait so a future reordering degrades instead of deadlocking.

### Suggested fix

Have `RouterTestFramework` await confirmation that the group actually started before running the prelude/test body — e.g. an internal readiness signal, or (if swift-service-lifecycle grows one) `ServiceGroup`'s own start signal. Filed the complementary report at swift-server/swift-service-lifecycle#248, since arguably `ServiceGroup` shouldn't silently discard an entire start either.

Happy to open a PR if you'd like this fixed here.

🤖 Analysis and reproduction assisted by [Claude Code](https://claude.com/claude-code)

adam-fowler:
Fixes #871 
We wait until a stream ends before running test closure. Add a service that finishes the stream, when service runs (indicating service group is running). Further down code wait until stream finishes.

### DIFF
--- a/Sources/HummingbirdTesting/RouterTestFramework.swift
+++ b/Sources/HummingbirdTesting/RouterTestFramework.swift
@@ -60,9 +60,10 @@ struct RouterTestFramework<Responder: HTTPResponder>: ApplicationTestFramework w
         }
         // if we have services then setup task group with service group running in separate task from test
         return try await withThrowingTaskGroup(of: Void.self) { group in
+            let (stream, cont) = AsyncStream.makeStream(of: Void.self)
             let serviceGroup = ServiceGroup(
                 configuration: .init(
-                    services: self.services,
+                    services: self.services + [FinishContinuationService(cont: cont)],
                     gracefulShutdownSignals: [.sigterm, .sigint],
                     logger: self.logger
                 )
@@ -71,12 +72,18 @@ struct RouterTestFramework<Responder: HTTPResponder>: ApplicationTestFramework w
                 try await serviceGroup.run()
             }
             do {
-                // run the runBeforeServer processes before we run test closure. Need to do this
-                // after we have run the serviceGroup though
+                // Before we run the test closure we need to run the runBeforeServer processes and also
+                // wait for the ServiceGroup to actually start. If we don't wait for the service group to
+                // start and the test closure returns immediately there is a possibility we call triggerGracefulShutdown
+                // before we actually run the service group and it will never run the services in the
+                // group. See https://github.com/hummingbird-project/hummingbird/issues/871 for details
                 for process in self.processesRunBeforeServerStart {
                     try await process()
                 }
+                await stream.first { _ in true }
+
                 let value = try await test(client)
+
                 await serviceGroup.triggerGracefulShutdown()
                 return value
             } catch {
@@ -161,6 +168,14 @@ struct RouterTestFramework<Responder: HTTPResponder>: ApplicationTestFramework w
                 values.trailingHeaders = headers
             }
         }
+    }
+}
+
+private struct FinishContinuationService: Service {
+    let cont: AsyncStream<Void>.Continuation
+    func run() async throws {
+        cont.finish()
+        try? await gracefulShutdown()
     }
 }
 


Sources/HummingbirdTesting/RouterTestFramework.swift
```diff
@@ -60,9 +60,10 @@ struct RouterTestFramework<Responder: HTTPResponder>: ApplicationTestFramework w
         }
         // if we have services then setup task group with service group running in separate task from test
         return try await withThrowingTaskGroup(of: Void.self) { group in
+            let (stream, cont) = AsyncStream.makeStream(of: Void.self)
```
Joannis, line 63: Wouldn't a `CheckedContinuatin` make more sense than a stream?

adam-fowler: How would you get a continuation into the Service. I don't want to wait on it immediately.

Joannis: Maybe I'm wrong, but I'm thinking of this:

```swift
var services = services
async let finished = await withCheckedContinuation { continuation in
    services.append(FinishContinuationService(cont: cont))
}

// Run services
await finished
```

adam-fowler: You can't guarantee that the child task that runs the `withCheckedContinuation` will run before calling `ServiceGroup.run()`

adam-fowler: In actual fact the child task ends of editing a copy of the services array, if it wasn't a copy it would be a Sendable error

Joannis (approved):

PR #889, opened 2026-08-18
Make ResponseBodyWriter methods nonisolated(nonsending)

adam-fowler:
This is a minor breaking change, but it allows for the ResponseBodyWriter to be used in `nonisolated(nonsending)` functions without a `Sending 'writer' risks causing data races` error. 

Without this many custom Response writers will struggle to use the ResponseBodyWriter.


adam-fowler:
This needs to wait for us to drop 6.1

PR #891, opened 2026-09-04
swift-http-server: Non-Copyable Request with borrowing responders

adam-fowler:


adam-fowler:
Closing for the moment, as a non-copyable structure would need to be mutated when ever the request body is consumed, which means we'd have to pass it as an inout parameter. Not great :-(

PR #899, opened 2026-09-18
Add `FullFoundation` trait that enables code needing the full Foundation library

adam-fowler:
This trait is enabled by default. So as long as people aren't setting traits when including hummingbird this change is non-breaking. But if someone was to set traits to something other than `.default` then the DateFormatter from URLEncodedFormEncoder/Decoder will not be available.

This PR also comes with an additional GitHub action that verifies whether the library links Foundation

### DIFF
--- a/.github/workflows/verify-foundation.yml
+++ b/.github/workflows/verify-foundation.yml
@@ -1,0 +1,22 @@
+name: Verify not using Foundation
+
+on:
+  pull_request:
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}-foundation
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  linux:
+    runs-on: ubuntu-latest
+    timeout-minutes: 15
+    container:
+      image: swift:latest
+    steps:
+    - name: Checkout
+      uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1  # v7.0.1
+    - name: run script
+      run: ./scripts/verify-foundation.sh
--- a/scripts/verify-foundation.sh
+++ b/scripts/verify-foundation.sh
@@ -1,0 +1,45 @@
+#!/bin/bash
+# Verify package with no trait does not require libFoundation.so or libFoundationInternationalization.so
+# This script only checks the dynamic libs referenced by the PerformanceTest executable.
+
+log() { printf -- "** %s\n" "$*" >&2; }
+error() { printf -- "** ERROR: %s\n" "$*" >&2; }
+fatal() { error "$@"; exit 1; }
+
+
+PRODUCT=$1
+TRAITS=${2:-}
+
+echo "Build $PRODUCT"
+if [ -z "$TRAITS" ]; then
+    swift build -c release --disable-default-traits --product "$PRODUCT"
+else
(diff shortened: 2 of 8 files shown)


.github/workflows/verify-foundation.yml
```diff
@@ -0,0 +1,22 @@
+name: Verify not using Foundation
```
tib:
```suggestion
- name: Verify that Foundation is not used
```

scripts/verify-foundation.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+# Verify package with no trait does not require libFoundation.so or libFoundationInternationalization.so
```
tib:
```suggestion
# Verify that the package without traits does not require libFoundation.so or libFoundationInternationalization.so
```

scripts/verify-foundation.sh
```diff
@@ -0,0 +1,45 @@
...
+OBJDUMP=$(objdump -p "$BINPATH"/"$PRODUCT" | grep "NEEDED")
+
+LIBS_TO_CHECK="libFoundation.so libFoundationInternationalization.so lib_FoundationICU.so"
+for LIB in ${LIBS_TO_CHECK}; do
+  echo -n "Checking for ${LIB}... "
+  
+  # check if the binary has a dependency on Foundation or ICU
+  echo "${OBJDUMP}" | grep "${LIB}"  # return 1 if not found
+
+  # 1 is success (grep failed to find the lib), 0 is failure (grep successly found the lib)
```
tib:
```suggestion
  # 1 is success (grep failed to find the lib), 0 is failure (grep successfully found the lib)
```

scripts/verify-foundation.sh
```diff
@@ -0,0 +1,45 @@
...
+  else
+    log "✅ ${LIB} not found."
+  fi
+done
+
+# exit code is the opposite of the grep exit code
+if [ "$SUCCESS" -eq 0 ]; then
+  fatal "❌ At least one foundation lib was found, reporting the error."
+else
+  log "✅ No foundation lib found, congrats!" && exit 0
```
tib:
```suggestion
  log "✅ No Foundation libraries found." && exit 0
```

tib (approved):
Found a few minor grammatical issues. Feel free to accept or ignore them, otherwise this looks good to me. 👍

the-braveknight:
Any ETA for when this is to going to be merged? I was just going to file a PR for linking the conditional linking of FoundationEssentials when available and noticed this branch. Also, out of curiosity, why use a trait when we can go only with the more modern FoundationEssentials APIs like FormatStyle, ParseStrategy, etc?

adam-fowler:
Removing the old Foundation APIs is a breaking change. In theory this PR is a breaking change as well given someone including HB with no traits will lose the foundation APIs but given they can just add the FullFoundation trait I think that's ok

the-braveknight:
> Removing the old Foundation APIs is a breaking change. In theory this PR is a breaking change as well given someone including HB with no traits will lose the foundation APIs but given they can just add the FullFoundation trait I think that's ok

How would it be a breaking change since the public API and the output results will be identical for consumers?

adam-fowler:
The public API includes DateFormatter which is part of the full foundation. If I remove those APIs any project dependent on them will stop compiling.

